PR #52, opened 2021-12-29
Support for c++17

sjunges:
On my machine I have switched to c++17. 

I would suggest to merge this pull request after the next release and require c++17 from that point onwards.


volkm (approved):
LGTM

PR #60, opened 2022-05-18
Remove python installation in CI

volkm:
Package `python` is not available under Ubuntu 22.04 any more.


volkm:
CI failure is due to more strict checks in Storm. Will merge.

PR #61, opened 2022-05-24
first version of documentation for simulators

sjunges:


volkm:
The test on the Jupyter notebook fails as the output is different than expected. Maybe the output is not deterministic?

sjunges:
Something got lost here, will redo.

PR #66, opened 2022-07-18
Dockerfile for stormpy

volkm:
Created Dockerfile for stormpy.
(Also removed obsolete files from Travis.)


.dockerignore
```diff
@@ -0,0 +1,21 @@
+Dockerfile
```
sjunges: I always like to have these kind of Dockerfile in the container for reference of people that only see the container.

sjunges:
Feel free to ignore the comment if you want.

LGTM, many thanks!!

volkm:
> Feel free to ignore the comment if you want.

I would never ignore your comments ;).

Thanks for the review. I will merge after the CI was successful.

PR #70, opened 2022-07-27
basic support for multiobjective model checking, including plotting

sjunges:


volkm:
The changes look good to me in general. As the code requires some dependencies (numpy, matplotlib, scipy?), I suggest to add them as optional dependencies to `setup.py` and add the necessary decorators for the tests. We already do this for numpy for example.

volkm:
Thanks, looks good. The test for matplotlib is currently skipped.
Do we want to also add a new `extra` in [setup.py](htps://github.com/moves-rwth/stormpy/blob/master/setup.py#L280)?

volkm:
conversation continues in https://github.com/moves-rwth/stormpy/issues/72

Issue #72, opened 2022-07-28
Matplotlib optional dependency


sjunges:
There is (very limited) functionality to support plotting in stormpy; which I like, but I dont want people to force installing matplotlib. Should we put this dependency into extras? It is also debatable whether CI should have a version with / without matplotlib?

volkm:
Yes, I think we should have `plotting` in `extras` for users which want to have this functionality.

For testing, I think we should test two versions: one with all extras and the minimum version.

volkm:
Closed with #73

PR #73, opened 2022-07-29
Optional dependencies for plotting

volkm:
- Added option `plot` to `setup.py`.
- Added in depth CI tests to build full functionality with all dependencies


volkm:
Addresses #72

PR #80, opened 2022-08-01
Improved Binder support

volkm:
- Updated Dockerfile for Binder to also include `examples`
- Updated `.dockerignore`
- Added badge to try out stormpy via JupyterLab


sjunges:
LGTM

Issue #68, opened 2022-07-19
Update Pybind


sjunges:
Will resolve the security PR https://github.com/moves-rwth/stormpy/pull/63 that changes pybind.

volkm:
Updated pybind to version 2.10 via PR #86.

PR #86, opened 2022-08-31
Download pybind11 via CMake instead of shipping it

volkm:
Download pybind11 in the configure step such that it is available for `add_subdirectory`.
Should best be merged simultaneous to the similar PR in pycarl https://github.com/moves-rwth/pycarl/pull/15/.


volkm:
Will also update pybind to version 2.10.0 and should fix #68.

sjunges:
I think that ultimately, it would be good if we could set the pybind version as an option. That way, if someone depends on a particular pybind version, it is easier to fix. Not urgent though. 

Otherwise: LGTM and many thanks

volkm:
One can already define the version in `include_pybind11.cmake`. Setting it in CMake and from Python `setup.py` should be easy to implement. Setting the version should however be an expert option, because pybind updates might require code adaptions.

sjunges:
As it currently must be the same as pycarls version, does it make sense to take the pycarl version as default?

volkm:
The recent changes allow to set the pybind version to use. As default we are using the pybind version exported by pycarl (see https://github.com/moves-rwth/pycarl/pull/15). Note that this requires to import pycarl during the build process and pycarl was added to `setup_requires`.

volkm:
Tests are currently failing, because https://github.com/moves-rwth/pycarl/pull/15 is not used yet.

sjunges:
LGTM

PR #88, opened 2022-09-01
Improve build process

volkm:
(Similar to the changes in pycarl PR https://github.com/moves-rwth/pycarl/pull/16)

- Configure only once at the start instead of configuring for each extension. This should slightly improve build performance.
- Let CMake generate the config files instead of `setup.py`.
- Both changes make the subdir for each extension obsolete.
- Created CMake macros for checking availability of libraries, etc.
- Also added CI tests for core functionality without any additional libraries
- Fixed flag `--disable-pomdp`


setup.py
```diff
@@ -108,92 +107,46 @@ def run(self):
         else:
             print("Stormpy - Warning: No support for POMDP analysis!")
 
-        # Set general cmake build options
         build_type = 'Debug' if self.config.get_as_bool("debug") else 'Release'
-        cmake_args = ['-DPYTHON_EXECUTABLE=' + sys.executable]
+        # Set cmake build options
+        print(self._extdir("core"))
```
sjunges: Is this print intended? If yes, could you add some text around it?

sjunges:
This is excellent; perfect! (See the comment about the print though)

setup.py
volkm: Good catch. That is a left-over from some debugging. Will remove it.

PR #94, opened 2022-09-06
Replaced deprecated distutils.Version

### LINKED ISSUE #93: Deprecation warning
I get the following warning when running stormpy on python 3.9.x

```
/Users/junges/stormpy/setup.py:80: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
  if StrictVersion(storm_version) < StrictVersion(storm_min_version):
```

volkm:
Fix #93

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,6 @@ import datetime
 
 from setuptools import setup, Extension, find_packages
 from setuptools.command.build_ext import build_ext
-from packaging.version import Version
 
 import setup.helper as setup_helper
 from setup.config import SetupConfig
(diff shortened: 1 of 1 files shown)


sjunges:
That was quick... Is this a solution for older python versions? Do we need to be careful?

volkm:
... and it fails. I should add the package as requirement.

To answer your question: `packaging` requires at least Python 3.7. Ubuntu 18.04 comes with Python 3.6. So we might need a distinction for older Python versions.

sjunges:
I think we should ditch 18.04. I dont want to write additional code for versions that old...

volkm:
We require at least Python version 3.7 now. The Storm Docker containers are all based on Ubuntu 22.04, so the tests should be fine.

sjunges:
LGTM

PR #99, opened 2022-09-18
Bindings for DFT modules

volkm:
- Adapted bindings for DFT modules (cf. https://github.com/moves-rwth/storm/pull/278)
- Export of modules to JSON format


sjunges:
LGTM

PR #100, opened 2022-10-02
Fix for DFT modules

volkm:
Fix according following changes in https://github.com/moves-rwth/storm/pull/305


volkm:
Tests are running again.

PR #103, opened 2022-10-19
Support for has_cycle() on sparse models

volkm:
Checking whether a model is acyclic should be quite helpful.

Note that the current implementation first needs to find and exclude all states with self-loops, because the underlying Storm function is considering self-loops as cycles.

Thanks to @jipspel for some help.

### DIFF
--- a/src/storage/model.cpp
+++ b/src/storage/model.cpp
@@ -15,11 +15,10 @@
 #include "storm/models/symbolic/Ctmc.h"
 #include "storm/models/symbolic/MarkovAutomaton.h"
 #include "storm/models/symbolic/StandardRewardModel.h"
-#include "storm/utility/dd.h"
-
 #include "storm/storage/dd/DdManager.h"
-
 #include "storm/storage/Scheduler.h"
+#include "storm/utility/dd.h"
+#include "storm/utility/graph.h"
 
 #include <functional>
 #include <string>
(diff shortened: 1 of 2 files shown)


src/storage/model.cpp
```diff
@@ -90,6 +89,26 @@ storm::storage::BitVector statesWithParameter(SparseModel<RationalFunction> cons
...
+                    // Found self-loop
+                    subStates.set(entry.getColumn(), false);
+                }
+            }
+            if (entry.getColumn() >= group) {
+                continue;
+            }
+        }
+    }
+    return storm::utility::graph::hasCycle(matrix, subStates);
```
tquatmann, line 110: This is problematic for MDPs, when  we have one self-loop row first and then one non-selfloop row which induces a cycle with other states, right?

I think the "hasCycle" would need to have a "subChoices" input to make this work?
Alternatively, this method could also build an SCC decomposition and we then check if each SCC has size 1.

volkm: You are right, this is indeed problematic.
I am open for other suggestion. The SCC decomposition might be a good idea.
I actually just wanted to have a simple function to check for acyclity. Unfortunately, `hasCycle` is not exactly giving me what I want. Maybe it makes sense to implement a better function in Storm directly.

sjunges:
Shall we close this PR?

volkm:
Yes, we can close it.

PR #112, opened 2023-03-28
Adaption to Storm changes

volkm:
Replaced one instance of `boost::optional` with `std::optional` to adapt to https://github.com/moves-rwth/storm/pull/340.
More replacements will be needed if Storm replaces all `boost::optional`.


sjunges:
LGTM, many thanks!

PR #113, opened 2023-04-20
Fix parametric models doc ipynb to use pdtmc

lisaoakley:
The parametric models documentation ipynb was not using the pDTMC die file, so the model used in the example was not parametric. I fixed this and fixed the type of the point assignments to match the linked example python file. The notebook runs correctly now until the constraint collector cell.

This exposes an issue later on in the ipynb, however. When the constraint collector is called on the correctly typed model (pDTMC model), an error is thrown that I am struggling to parse. I'm not sure where or how to fix this, but it seems to be an orthogonal problem to the issues in the earlier cells, so it should be ok to merge these changes and then open an issue for this bug separately.

Error message about the constraints collector portion here:

```---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
TypeError: Unregistered type : carl::Formula<carl::MultivariatePolynomial<cln::cl_RA, carl::MonomialComparator<&(carl::Monomial::compareGradedLexical(std::__1::shared_ptr<carl::Monomial const> const&, std::__1::shared_ptr<carl::Monomial const> const&)), true>, carl::StdMultivariatePolynomialPolicies<carl::NoReasons, carl::NoAllocator> > >

The above exception was the direct cause of the following exception:

TypeError                                 Traceback (most recent call last)
Cell In[7], line 2
      1 collector = stormpy.ConstraintCollector(model)
----> 2 for formula in collector.wellformed_constraints:
      3     print(formula)
      4 for formula in collector.graph_preserving_constraints:

TypeError: Unable to convert function return value to a Python type! The signature was
	(arg0: stormpy.core.ConstraintCollector) -> Set[carl::Formula<carl::MultivariatePolynomial<cln::cl_RA, carl::MonomialComparator<&(carl::Monomial::compareGradedLexical(std::__1::shared_ptr<carl::Monomial const> const&, std::__1::shared_ptr<carl::Monomial const> const&)), true>, carl::StdMultivariatePolynomialPolicies<carl::NoReasons, carl::NoAllocator> > >]

Did you forget to `#include <pybind11/stl.h>`? Or <pybind11/complex.h>,
<pybind11/functional.h>, <pybind11/chrono.h>, etc. Some automatic
conversions are optional and require extra headers to be included
when compiling your pybind11 module.
```

### DIFF
--- a/doc/source/doc/parametric_models.ipynb
+++ b/doc/source/doc/parametric_models.ipynb
@@ -30,7 +30,7 @@
     ">>> import stormpy\n",
     ">>> import stormpy.examples\n",
     ">>> import stormpy.examples.files\n",
-    ">>> path = stormpy.examples.files.prism_dtmc_die\n",
+    ">>> path = stormpy.examples.files.prism_pdtmc_die\n",
     ">>> prism_program = stormpy.parse_prism_program(path)\n",
     ">>> formula_str = \"P=? [F s=7 & d=2]\"\n",
     ">>> properties = stormpy.parse_properties(formula_str, prism_program)\n",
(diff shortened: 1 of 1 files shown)


sjunges:
Hi Lisa,


many thanks, this looks good to me. I am working on refurbishing the parametric API, and will be able to fix the additional problem as I go.

volkm:
The CI tests fail as expected, because of the remaining issue with the `ConstraintCollector`. We can merge it.
Thanks for the fix, Lisa.

PR #116, opened 2023-04-26
More robust multi-objective test.

tquatmann:


sjunges:
Does this model have any other vertices? I think it is fine as is, but one may need to adapt this in the future?

Sebastian

tquatmann:
The given set of expected vertices is the exact solution to the problem. It's a 3-state MDP so I was able to get those values. Any computed vertex should be sufficiently close. Theoretically, there could be some spurious vertices in the over-approximation, but those can't appear with the currently implemented algorithm.

sjunges:
Thanks!

PR #126, opened 2023-07-01
fix for boost 1.82

sjunges:
Mimicks https://github.com/moves-rwth/storm/pull/387


volkm:
Good idea.
Not to be nit-picky, but can we maybe use `if (${Boost_VERSION} VERSION_GREATER_EQUAL "1.81.0")` and increase the required CMake version to 3.15? Then we have the same state as Storm (see https://github.com/moves-rwth/storm/pull/393).

PR #133, opened 2023-09-20
Add bindings for computing steady-state distributions

### LINKED ISSUE #132: Computing steady-state probability vector
Hello everyone,

thank you for this tool, we use it on version 1.7.0, and it has been great for our purpose, computing the steady-state probability for a set of states in a ctmc. However, working on some more complex models recently, we expect to run into computation time issues when checking the desired property directly. Hence searching for ways to reduce the size of the state space, we came across the technique of lumping. Although we are still in the process of determining our detailed approach, we are considering a solution that requires more 'manual' handling of steady-state probabilities to modify transition rates in sequentially solved ctmcs. Since we would love to start the sequential solving utilizing stormpy, we are searching for a way to obtain the steady-state probabilities for all states in a considered ctmc.  

Unfortunately, we have not been able to find an example describing how to obtain the desired results. Is there a method that does provide access to all steady-state probabilites of the considered state space?

Thanks in advance :)

tquatmann:
fixes #132

### DIFF
--- a/tests/core/test_modelchecking.py
+++ b/tests/core/test_modelchecking.py
@@ -179,3 +179,16 @@ class TestModelChecking:
         result = stormpy.compute_expected_number_of_visits(environment, model)
         assert result.at(0) == 1
         assert math.isclose(result.at(1),2.0/3)
+        
+    def test_compute_steady_state_distributioin(self):
+        program = stormpy.parse_prism_program(get_example_path("dtmc", "die.pm"))
+        model = stormpy.build_model(program)
+        environment = stormpy.Environment()
+        result = stormpy.compute_steady_state_distribution(environment, model)
+        values = result.get_values()
+        assert len(values) == 13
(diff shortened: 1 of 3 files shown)


tests/core/test_modelchecking.py
```diff
@@ -179,3 +179,16 @@ def test_compute_expected_number_of_visits(self):
         result = stormpy.compute_expected_number_of_visits(environment, model)
         assert result.at(0) == 1
         assert math.isclose(result.at(1),2.0/3)
+        
+    def test_compute_steady_state_distributioin(self):
```
volkm: Small typo "distribution"

volkm (approved):
Looks good! Thanks

tests/core/test_modelchecking.py
tquatmann: Good catch!  Fixed!

PR #137, opened 2023-10-31
CI: improved build configurations

volkm:
- Use correct Storm Docker image `ci` instead of `ci-release`.
- Replaced environment variables by configuration variables
- Updated some Github actions to latest version


sjunges:
LGTM (this is indeed much cleaner now) :-)

volkm:
Thanks, I also think that it is cleaner now.

PR #144, opened 2023-11-27
Patch to pybind11 to avoid issue with concat

### LINKED ISSUE #140: Build fails after changes in modernjson
The update of the modernjson library (see https://github.com/moves-rwth/storm/pull/424) leads to build failures:
```
error: constexpr variable cannot have non-literal type 'const std::string'
```

Adding the [pybind11_json](https://github.com/pybind/pybind11_json) header file seems to resolve the problem for ValueType double, but not for RationalNumber.

A more general question could be whether we need Python bindings for the `storm::json` object or whether it would suffice to directly export json as string within C++.

volkm:
Add namespace `detail::concat` in `pybind11/include/pybind11/cast.h`. This resolves https://github.com/moves-rwth/stormpy/issues/140 for now.
We might want to create a better solution in the future.

### DIFF
--- a/resources/pybind11_2.10.0.patch
+++ b/resources/pybind11_2.10.0.patch
@@ -1,0 +1,22 @@
+diff --git a/include/pybind11/cast.h b/include/pybind11/cast.h
+index a0e32281..f132d096 100644
+--- a/include/pybind11/cast.h
++++ b/include/pybind11/cast.h
+@@ -657,7 +657,7 @@ public:
+     }
+ 
+     static constexpr auto name
+-        = const_name("Tuple[") + concat(make_caster<Ts>::name...) + const_name("]");
++        = const_name("Tuple[") + pybind11::detail::concat(make_caster<Ts>::name...) + const_name("]");
+ 
+     template <typename T>
+     using cast_op_type = type;
+@@ -1399,7 +1399,7 @@ public:
+     static_assert(args_pos == -1 || args_pos == constexpr_first<argument_is_args, Args...>(),
+                   "py::args cannot be specified more than once");
+ 
+-    static constexpr auto arg_names = concat(type_descr(make_caster<Args>::name)...);
++    static constexpr auto arg_names = pybind11::detail::concat(type_descr(make_caster<Args>::name)...);
+ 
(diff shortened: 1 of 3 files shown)


resources/pybind11_2.10.0.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/include/pybind11/cast.h b/include/pybind11/cast.h
+index a0e32281..f132d096 100644
+--- a/include/pybind11/cast.h
++++ b/include/pybind11/cast.h
+@@ -657,7 +657,7 @@ public:
+     }
+ 
+     static constexpr auto name
+-        = const_name("Tuple[") + concat(make_caster<Ts>::name...) + const_name("]");
++        = const_name("Tuple[") + detail::concat(make_caster<Ts>::name...) + const_name("]");
```
tquatmann:
```suggestion
+        = const_name("Tuple[") + pybind11::detail::concat(make_caster<Ts>::name...) + const_name("]");
```
This seems more robust to me since the other `concat` also lies in a `detail` namespace.

resources/pybind11_2.10.0.patch
```diff
@@ -0,0 +1,22 @@
...
+ 
+     template <typename T>
+     using cast_op_type = type;
+@@ -1399,7 +1399,7 @@ public:
+     static_assert(args_pos == -1 || args_pos == constexpr_first<argument_is_args, Args...>(),
+                   "py::args cannot be specified more than once");
+ 
+-    static constexpr auto arg_names = concat(type_descr(make_caster<Args>::name)...);
++    static constexpr auto arg_names = detail::concat(type_descr(make_caster<Args>::name)...);
```
tquatmann:
```suggestion
+    static constexpr auto arg_names = pybind11::detail::concat(type_descr(make_caster<Args>::name)...);
```

tquatmann (approved):
Thanks for the fix!
I'm curious why the wrong concat function is used, as it's supposed to be in another namespace (nlohmann).
However, feel free to merge this.

resources/pybind11_2.10.0.patch
volkm: Good point, thanks.

PR #146, opened 2023-11-29
currently no sysversions of pybind

sjunges:
As we patch pybind, we currently cannot rely on system versions.

Btw, are we planning to submit a patch to pybind?


volkm:
Good point. LGTM.

volkm:
I thought about submitting a patch to pybind. I did not see anybody yet having the same problem.
I first have to setup the project and run the tests, but hopefully it should be smooth.

PR #150, opened 2023-12-19
Bindings for intervals

volkm:
Support for intervals introduced with https://github.com/moves-rwth/storm/pull/409

### DIFF
--- a/lib/stormpy/__init__.py
+++ b/lib/stormpy/__init__.py
@@ -216,6 +216,22 @@ def build_parametric_model_from_drn(file, options = DirectEncodingParserOptions(
     return _convert_sparse_model(intermediate, parametric=True)
 
 
+def build_interval_model_from_drn(file, options = DirectEncodingParserOptions()):
+    """
+    Build an interval model in sparse representation from the explicit DRN representation.
+
+    :param String file: DRN file containing the model.
+    :param DirectEncodingParserOptions: Options for the parser.
+    :return: Interval model in sparse representation.
+    """
+    intermediate = core._build_sparse_interval_model_from_drn(file, options)
+    assert intermediate.supports_uncertainty
+    if intermediate.model_type == ModelType.MDP:
+        return intermediate._as_sparse_imdp()
+    else:
+        raise StormError("Not supported interval model constructed")
+
+
 def perform_bisimulation(model, properties, bisimulation_type):
     """
     Perform bisimulation on model.
--- a/lib/stormpy/__init__.py
+++ b/lib/stormpy/__init__.py
@@ -300,7 +300,9 @@ def check_model_sparse(model, property, only_initial_states=False, extract_sched
         if force_fully_observable:
             # Note that casting a model to a fully observable model wont work with python/pybind, so we actually have other access points
             if model.supports_parameters:
-                raise NotImplementedError("")
+                raise NotImplementedError("Model checking of partially observable models is not supported for parametric models.")
+            elif model.supports_uncertainty:
+                raise NotImplementedError("Model checking of partially observable models is not supported for interval models.")
             elif model.is_exact:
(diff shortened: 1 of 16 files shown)


lib/stormpy/__init__.py
```diff
@@ -300,7 +300,9 @@ def check_model_sparse(model, property, only_initial_states=False, extract_sched
         if force_fully_observable:
             # Note that casting a model to a fully observable model wont work with python/pybind, so we actually have other access points
             if model.supports_parameters:
-                raise NotImplementedError("")
+                raise NotImplementedError("Model checking of partially observable models is not supported for parametric models.")
+            elif model.supports_uncertainty:
+                raise NotImplementedError("Model checking of partially observable models is not supported for interval models.")
```
sjunges, line 321: but this is if force_fully_observable is set to true, which is basically casting the POMDP to an MDP. It is debatable whether this part of the API is pretty though

lib/stormpy/__init__.py
```diff
@@ -600,11 +602,11 @@ def parse_properties(properties, context=None, filters=None):
         if context.is_prism_program():
             return core.parse_properties_for_prism_program(properties, context.as_prism_program(), filters)
         else:
-            core.parse_properties_for_jani_program(properties, context.as_jani_model(), filters)
+            return core.parse_properties_for_jani_program(properties, context.as_jani_model(), filters)
```
sjunges, line 621: 👍

src/core/modelchecking.cpp
```diff
@@ -46,6 +49,12 @@ std::shared_ptr<storm::modelchecker::CheckResult> modelCheckingHybridEngine(std:
     return storm::api::verifyWithHybridEngine<DdType, ValueType>(env, model, task);
 }
 
+std::shared_ptr<storm::modelchecker::CheckResult> checkIntervalMdp(std::shared_ptr<storm::models::sparse::Mdp<storm::Interval>> mdp, CheckTask<double> const& task, storm::Environment& env) {
+    env.solver().minMax().setMethod(storm::solver::MinMaxMethod::ValueIteration);
```
sjunges: I would prefer not to magically change the environment and have the user be responsible for this. Also because I hope to soon have extensions to more standard settings.

src/storage/scheduler.cpp
```diff
@@ -21,32 +21,44 @@ void define_scheduler(py::module& m, std::string vt_suffix) {
...
+            // Conversion from RationalFunction to Interval not implemented
+            scheduler.def("cast_to_interval_datatype", &Scheduler::template toValueType<storm::Interval>, "Construct the scheduler with `interval` value type");
+        }
+    }
+
+
     std::string schedulerChoiceClassName = std::string("SchedulerChoice") + vt_suffix;
     py::class_<SchedulerChoice> schedulerChoice(m, schedulerChoiceClassName.c_str(), "A choice of a finite memory scheduler");
     schedulerChoice
+        .def(py::init<uint_fast64_t>(), "choice"_a)
```
sjunges: uint64_t

src/core/modelchecking.cpp
volkm: We have bindings to change the environment from stormpy which we can use here. Users then will have to make sure to use the right solver for intervals (at least for now).

lib/stormpy/__init__.py
volkm: That is a good point. To keep things a bit clear, a opened a new issue, see https://github.com/moves-rwth/stormpy/issues/151. My proposal is to keep it here for now but change it in a separate PR.

sjunges: ok but the error message is confusing/misplaced :-)

sjunges:
LGTM :-)

PR #154, opened 2024-01-02
some ADD support, improved valuation support, towards access for resu…

sjunges:
with some ADD support, improved valuation support


volkm (approved):
LGTM.
If time permits, maybe add some tests at some point, because iterators can sometimes behave a bit unexpected.

PR #158, opened 2024-02-14
Add interpreter lock release to POMDP check function

AlexBork:
The POMDP check function binding is lacking the release of the global interpreter lock that allows other processes to take control of a stopped belief exploration model checker.

Thanks to @TheGreatfpmK for pointing this out.


volkm:
Can you elaborate a bit more on the change? As far as I understand, `py::gil_scoped_release>()` releases the Global Interpreter Lock and therefore potentially allows other threads to access the model checker. From your description you want to achieve the opposite though?

AlexBork:
Sorry, there was a misunderstanding of the problem on my side. The problem was indeed that the model checker was keeping the interpreter lock, so other processes were not able to restart it while it was suspended. I have changed the title and description to account for that.

volkm:
Thanks for the clarification.
We have to be careful when releasing the GIL to avoid race conditions with multiple threads. I am okay with the change for now, but maybe we should think about general handling of parallelization in the future.
Any opinion, @sjunges?

sjunges:
Sorry, I was not aware of this issue. I am also concerned about this, in particular as it depends on the options whether or not this is actuallly code that is meant to run in parallel.

I would be more comfortable with a function run_check  or something. We could then have a more general policy that such functions that are meant to run in parallel are always starting with run_*?

volkm:
Maybe [checking for signals in pybind](https://pybind11.readthedocs.io/en/stable/faq.html#how-can-i-properly-handle-ctrl-c-in-long-running-functions) is an option?

sjunges:
Is there any update on this? I don't think we will merge this code here as is. Marking it for closed right now, happy to reopen though.

PR #173, opened 2024-07-13
Add bindings for the Smg class

TheGreatfpmK:
Added all the needed bindings for the Smg class.

Also fixed SparseModelComponents binding wrong property as 'player1_matrix'.


sjunges:
LGTM. Is there a way to create an SMG inside stormpy, e.g., via the components? Would be great if you could add a small test case for this to ensure that we do not break things in the future.  (Maybe in tests/storage/...)?

TheGreatfpmK:
I haven't tried building the model in StormPy, but I will take a look! I will also add a test.

TheGreatfpmK:
I added a small test case for constructing SMG from sparse model components!

sjunges:
Great. Once the tests finish, I will merge it!

PR #174, opened 2024-07-15
Adaption to changes in Storm

volkm:
Tests failed due to changes in https://github.com/moves-rwth/storm/pull/556

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -14,8 +14,6 @@ on:
   pull_request:
 
 env:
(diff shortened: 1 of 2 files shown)


sjunges:
Many thanks! LGTM. But should there be a default argument here?

volkm:
The original Storm methods do not have a default either. Is `true` as sensible default value?

sjunges:
Probably it is but if the C++ methods do not have a default, then it is maybe good to not introduce one here right now :-)

volkm:
Then I will keep it unchanged for now.

PR #181, opened 2024-07-24
Add binding for GameFormula

TheGreatfpmK:
Added binding for the GameFormula class and one small test for it.


volkm:
Looks good, thanks!

PR #198, opened 2024-12-16
Bugfix in parse_properties

sjunges:
Minor bugfix: symbolic_description.is_prism_program is a property, not a method.


sjunges:
( I will of course stash before pushing... )

volkm (approved):
LGTM.
Stashing is indeed a good idea.

PR #199, opened 2024-12-22
Added convenience functions for state valuations

volkm:
Requires https://github.com/moves-rwth/storm/pull/650
- Replaced `program.variables` by `program.variables()` to follow changes in Storm
- Support for getting the value of a variable in all states
- Extended and added tests


volkm:
I opted to make `program.variables` a function which now takes an argument `include_constants`, similar to Storm. If we want to keep it as a member, we could change it by having two members `program.variables` (corresponding to `include_constants=False`) and `program.variables_and_constants` (`include_constants=True`).
@sjunges What is your view?

sjunges:
Should we call it get_variables() then? 

Also, I am wondering whether .variables should mimick the old behavior for some time? I am not sure at what point in time we should start deprecating things.

volkm:
Good idea, I will make it `get_variables()` and keep `variables` for the old behavior.

sjunges:
LGTM!

PR #208, opened 2025-02-18
PRISM SMGs can now be built through StormPy

TheGreatfpmK:
Fixed build of SMGs through StormPy.

I added an SMG test file and test for building SMGs.


volkm:
LGTM. Thanks for adding the test case.
I will merge it if the tests are successful.

PR #211, opened 2025-02-21
Update 'observations' to 'observability_classes'

marisgg:
For building POMDPs, the attribute name for setting the observations in the `stormpy.SparseModelComponents` class in Stormpy is `observability_classes`.

This change prevents readers' confusion following `building_mdps.ipynb` (like myself).


volkm:
Good catch. Thank for the contribution

PR #212, opened 2025-02-21
Generalized functions get_value and get_values_states for valuations

### LINKED ISSUE #210: Segfault in get_integer_values_states
Hi thanks for the library. I am trying to use the ```get_integer_values_states``` that was recently added to the library. 
I am using the master of Storm and of stormpy, since these functions were added after 1.9.0. I don't know if this is an issue of Storm or of the bindings. The types passed differ with the docs as prism_program.variables give a set of ```stormpy.storage.storage.Variable``` while ```get_integer_values_states``` seems to expect ```storm::expressions::Variable``` . I don't know if this might be a problem.  

```python
import stormpy

prism_program = stormpy.parse_prism_program(str(specification_path))  # type: ignore
options = stormpy.BuilderOptions()  # type: ignore
options.set_build_state_valuations()
options.set_build_choice_labels()

model = stormpy.build_sparse_model_with_options(prism_program, options)  # type: ignore

[
    model.state_valuations.get_integer_values_states(variable)
    for variable in prism_program.variables
]
```
You can repro with the simplest MDP:

```
mdp 

module test

a: bool;
b: bool;

[flip_a] (a=true) -> (a'=false);
[flip_a] (a=false) -> (a'=true);
[flip_b] (b=true) -> (b'=false);
[flip_b] (b=false) -> (b'=true);

endmodule
```

This gives: Segmentation fault (core dumped) after building the model.  Any help is appreciated.

volkm:
Fixes https://github.com/moves-rwth/stormpy/issues/210

Add general functions `get_value` and `get_values_states` which are type-independent and internally call the right Storm function.

### DIFF
--- a/examples/building_models/03-building-models.py
+++ b/examples/building_models/03-building-models.py
@@ -25,7 +25,7 @@ def example_building_models_03():
         print("module {}".format(module.name))
         integer_variables += module.integer_variables
 
-    print(", ".join(["{}: {}".format(str(iv.name), valuations.get_integer_value(2, iv.expression_variable)) for iv in integer_variables]))
+    print(", ".join(["{}: {}".format(str(iv.name), valuations.get_value(2, iv.expression_variable)) for iv in integer_variables]))
 
 
 if __name__ == "__main__":
--- a/examples/building_models/04-building-models.py
+++ b/examples/building_models/04-building-models.py
@@ -31,7 +31,7 @@ def example_building_models_04():
         # conditions on the action
         cond1 = action_name.startswith("snd")
         # conditions on the state (efficient)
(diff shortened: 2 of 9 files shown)


sjunges:
I agree this is better, but I wonder whether we already make get_integer_value and get_boolean_value private or whether we should first make them deprecated. This is one of these functions that may be used in other code by now?

volkm:
It is a general questions how backwards-compatible stormpy should be. I see that more people are using stormpy and we should try to limit the number of breaking changes.
I will add the old version as deprecated functions.

volkm:
A deprecated decorator is [introduced](https://docs.python.org/3.13/library/warnings.html#warnings.deprecated) by Python 3.13. As we also need to support older Python versions, I suggest to use the [Deprecated package](https://pypi.org/project/Deprecated/) for now.

sjunges:
LGTM! Many thanks

PR #219, opened 2025-04-07
Added support for building Interval POMDPs from drn

TheGreatfpmK:
Also added one example file for simple interval POMDP and added one test for building interval POMDP


TheGreatfpmK:
Not sure why the checks are failing here... I've tried and the newly added test works on 3 separate machines.

volkm:
I found the issue: in the debug mode the assertion [here](https://github.com/moves-rwth/storm/blob/master/src/storm/models/sparse/Model.cpp#L63) fails because not all entries are strictly positive. If you change the last entry of the drn file to e.g. `[0.1, 0.5]` it should work.

volkm:
Other than that, the contribution looks good to me and I will merge it as soon as the tests pass.

TheGreatfpmK:
Ah I see, thanks Matthias!

PR #225, opened 2025-05-27
Add argument for useGroups to subMatrix

### LINKED ISSUE #224: `submatrix` does not allow useGroups=false
In stormpy, the `submatrix` function fixes the first argument of Storm's `submatrix` method to `true`. This means that you cannot filter individual rows on MDPs.

https://github.com/moves-rwth/stormpy/blob/3536592dfba4e4d0504efc329e9ac2398b07cd07/src/storage/matrix.cpp#L97

linusheck:
This should fix https://github.com/moves-rwth/stormpy/issues/224


volkm:
Looks good. Can you add a test or change an existing one such that we test the flag?

sjunges:
I merged the fix by Matthias. If you merge the current master, the tests should run.

volkm (approved):
LGTM

PR #227, opened 2025-05-27
Added whitespace for POMDPs in DRN

volkm:
Required after change in https://github.com/moves-rwth/storm/pull/717/


volkm:
Is needed for Storm version 1.10.0 but can be reverted if https://github.com/moves-rwth/storm/issues/722 is addressed.

sjunges:
LGTM, not sure why it is needed though.

volkm:
Thanks.
The latest change on the DRN parser always requires a whitespace after an observation, even if only a linebreak follows. I think we should lift this requirement in Storm if there is a linebreak, see https://github.com/moves-rwth/storm/issues/722.

PR #229, opened 2025-06-01
Scikit for CmakeUpdates

### LINKED ISSUE #161: Package stormpy as a wheel
Hi,
it is easy to compile stormpy as a python wheel:

```
python3 -m build --wheel
=> Successfully built stormpy-1.8.0-cp312-cp312-macosx_14_0_arm64.whl

pip install ./dist/stormpy-1.8.0-cp312-cp312-macosx_14_0_arm64.whl
```

Why not package it as a wheel? The stormpy installation would go from compiling storm on the right version, compiling stormpy, etc., to just doing a `pip install stormpy`. There would need to be someone on a Linux and someone on a MacOS system compiling it every release, but I don't see this as a big issue.
You could also package it as a wheel under a different name or under a toggle, like `pip install stormpy[prebuilt]`.

### LINKED ISSUE #216: Get carl directory from Storm
Consider whether we can get the carl directory via storm (https://github.com/moves-rwth/storm/issues/62). This potentially also needs updates in (https://github.com/moves-rwth/storm/pull/602).

Follow up to point 2 of https://github.com/moves-rwth/stormpy/pull/205#issuecomment-2646555007

sjunges:
Updates that build upon the cmake updates in storm. This now allows building wheels.

Fixes https://github.com/moves-rwth/stormpy/issues/216.
Fixes https://github.com/moves-rwth/stormpy/issues/161


CMakeLists.txt
```diff
@@ -96,13 +114,20 @@ SET(CMAKE_SKIP_BUILD_RPATH  FALSE)
...
+endif()
+SET(CMAKE_INSTALL_RPATH "${RELPOS_STRING}/../../../lib/storm;${RELPOS_STRING}/../../../lib/storm/resources;${RELPOS_STRING}/../../lib/storm;${RELPOS_STRING}/../../lib/storm/resources;${RELPOS_STRING}/../lib/storm;${RELPOS_STRING}/../lib/storm/resources;${RELPOS_STRING}/lib/storm;${RELPOS_STRING}/lib/storm/resources;${RELPOS_STRING};${RELPOS_STRING}/resources")
+#  add the automatically determined parts of the RPATH
 # which point to directories outside the build tree to the install RPATH
 SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+SET(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
 
 
 # Workaround for issue with Boost >= 1.81
+# TODO is this still necessary?
```
volkm: We still have the same check in Storm, but we should indeed check whether this is still needed at some point.

pyproject.toml
```diff
@@ -9,10 +9,10 @@ readme = "README.md"
 license.file = "LICENSE"
 dynamic = ["version"]
 authors = [
-    { name = "S. Junges", email = "sjunges@cs.ru.nl" },
+    { name = "S. Junges", email = "sebastian.junges@ru.nl" },
     { name = "M. Volk", email = "m.volk@tue.nl" }
 ]
-requires-python = ">=3.7"
+requires-python = ">=3.11" # Can be reduced again, but want to iterate quicker through tests for building wheels.
```
volkm: Should be reduced again.
```suggestion
requires-python = ">=3.9"
```

src/pomdp/quantitative_analysis.cpp
```diff
@@ -1,8 +1,8 @@
 #include "quantitative_analysis.h"
 
-#include <storm/adapters/RationalFunctionAdapter.h>
-#include <storm/models/sparse/Pomdp.h>
-#include <storm-pomdp/api/verification.h>
+#include "storm/adapters/RationalFunctionAdapter.h"
+#include "storm/models/sparse/Pomdp.h"
+#include "storm-pomdp/api/verification.h"
```
volkm, line 5: Why is this change needed? Is using `<..>` not preferred?

.github/workflows/before-all-linux.sh
```diff
@@ -0,0 +1,34 @@
...
+tar -zxf sccache-v${sccache_version}-$(arch)-unknown-linux-musl.tar.gz
+cp sccache-v${sccache_version}-$(arch)-unknown-linux-musl/sccache /usr/bin
+
+## Install boost
+#curl -fsSLO https://archives.boost.io/release/${boost_version}/source/boost_${boost_version_under}.tar.bz2
+#tar -jxf boost_${boost_version_under}.tar.bz2
+## Quick-n-dirty approach (much faster than doing the install, which copies thousands of files)
+#ln -s /tmp/boost_${boost_version_under}/boost /usr/include/boost
+
+# Install ginac
```
volkm, line 27: Why can we not use the ginac package from the package manager?

volkm (commented):
Overall LGTM. I have some smaller clarifying questions.

Note that I opted to use pybind version 2.12.1 for now. Pybind versions 2.13 lead to a build error for me which I still need to investigate.

src/pomdp/quantitative_analysis.cpp
sjunges: I dont think so: https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

CMakeLists.txt
sjunges: I updated the comment to reflect a more nuanced version of this comment.

.github/workflows/before-all-linux.sh
sjunges: We probably can, but the general advice seems to be to not use brew (and similar things hold for  other package managers).

https://cibuildwheel.pypa.io/en/stable/faq/#missing-dependencies 

I think having the ability to build a version from source is probably good in terms of flexibility here.

src/pomdp/quantitative_analysis.cpp
volkm: Follow up in https://github.com/moves-rwth/stormpy/issues/231

.github/workflows/before-all-linux.sh
volkm: Ok, so we should try to not rely on homebrew. In this case, I suggest that we also try to build cln manually, and maybe some other libraries if needed.

PR #241, opened 2025-06-19
Use pybind from pip package

volkm:
I did not fix a specific pybind version for now, but we could add it if necessary.


sjunges:
Nice! LGTM!

PR #245, opened 2025-06-30
Revise stormpy modules

volkm:
- Add an underscore suffix to all pybind libraries, e.g. `_storage` instead of `storage`. This helps distinguishing between the Python module `storage` and the pybind library `_storage`. Some adaptions were required but nothing should change on the outside.
- Added new module `stormpy.exceptions` and added type information to docstring
- Revised documentation for `stormpy.info`. I tried making the functions `storm_version()`, etc properties, but this lead to issues because the property decorator is aimed at classes and not modules.
- Generate API documentation for `stormpy.pomdp`
- Use sphinx decorator for deprecation. Deprecations now require a version and I opted to use the last version `1.10.0`. Technically the deprecation was only introduced afterwards, but otherwise we need to reference a future version without knowing what version number we will actually have.
- Our Sphinx documentation generates documentation also for imported members (to include the pybind libraries). I needed to explicitly exclude the `deprecated` import as it would occur in the documentation otherwise. We have to keep an eye on future imports which we also might need to exclude.


sjunges:
LGTM!

PR #248, opened 2025-07-21
CI for publishing to Pypi

volkm:
Releases to Pypi can be manually triggered from the `wheelpypi.yml` workflow. Inputs are the storm version to use as base and whether to upload to test.pypi or to pypi.
The CI first builds Storm once and then builds stormpy for each Python version. This reduces build times compared to the previous version where Storm was built for each Python version. As an important side-effect, the wheels have also been reduced by factor 3. In the previous version, libspot was included thrice for some reason and significantly blew up the wheel size. This is not the case anymore.

The changes were successfully used to release the first stand-alone version of stormpy 1.10.1 to [pypi](https://pypi.org/project/stormpy/1.10.1/#files).


pyproject.toml
```diff
@@ -103,17 +103,15 @@ before-all = ".github/workflows/before-all-linux.sh"
...
 
 [tool.cibuildwheel.config-settings]
-"cmake.args" = ["-DSTORM_PORTABLE=ON", "-DALLOW_STORM_SYSTEM='OFF'"]
+"cmake.args" = ["-DSTORM_PORTABLE=ON", "-DALLOW_STORM_SYSTEM='ON'", "-DALLOW_STORM_FETCH='OFF'"]
 
 [tool.black]
 line-length = 160
 target-version = [
-    "py39",
-    "py310",
```
sjunges: Just a minor point: I think there is a question to be discussed whether python versions for which we ship and python versions that we support are the same per se?
In particular, I can imagine that we may want to support other python distributions (like conda) without shipping wheels.

sjunges:
Looks good and is properly tested ;-)

I would mention  in the PR description that we disabled Python 3.9,3.10 to avoid using too much space on pypi. 
There is a discussion whether that also means that we do not support these python versions.

I dont mind much right now.

pyproject.toml
volkm: I think we could enable Python 3.9 and 3.10 again for now. I only disabled it to speed up the build process and get smaller artifacts. Both issues should be mitigated now.

In general, I would like to build wheels for all supported Python version. Otherwise, it might get too confusing.

volkm:
Combined with https://github.com/moves-rwth/storm/pull/745 it resolves https://github.com/moves-rwth/stormpy/issues/249

volkm:
We support all Python versions >= 3.9 for now. I think it is good to have wheels for all currently supported Python versions to avoid issues for users with older systems. After the size reduction of the wheels, we should enough space for now.

PR #214, opened 2025-02-24
Fix nbsphinx version to avoid issue with version 8.2.0

volkm:
Fixes https://github.com/moves-rwth/stormpy/issues/213

PR #250, opened 2025-07-24
Use latest Sphinx version, reverts #214

volkm:
Reverts workaround introduced in #214.
Sphinx 8.2.2 is [fixed](https://github.com/spatialaudio/nbsphinx/issues/825#issuecomment-2692944098).


sjunges:
LGTM

PR #262, opened 2025-08-19
Update test_pomdp_quantitative_analysis.py

sjunges:
Be less precise in these results; this is not the intention of the test and they may cause failures.

### DIFF
--- a/tests/pomdp/test_pomdp_quantitative_analysis.py
+++ b/tests/pomdp/test_pomdp_quantitative_analysis.py
@@ -20,7 +20,7 @@ class TestPomdpQuantitative:
         options.use_clipping = False
         belmc = stormpy.pomdp.BeliefExplorationModelCheckerDouble(model, options)
         result = belmc.check(formulas[0].raw_formula, [])
-        assert math.isclose(result.lower_bound, 0.351985, abs_tol=10**-6)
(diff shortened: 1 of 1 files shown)


tests/pomdp/test_pomdp_quantitative_analysis.py
```diff
@@ -55,7 +55,7 @@ def test_underapprox_mc_cmaze_rmin(self):
         belmc = stormpy.pomdp.BeliefExplorationModelCheckerDouble(model, options)
         result = belmc.check(formulas[0].raw_formula, [])
         assert math.isinf(result.lower_bound)
-        assert math.isclose(result.upper_bound, 19.781437, abs_tol=10**-6)
+        assert math.isclose(result.upper_bound, 19.78127, abs_tol=10**-4)
```
volkm, line 58: Why is it a different number and not rounded to 19.78144?

volkm:
I agree that the result checking should be mainly performed in Storm and not stormpy. LGTM.

tests/pomdp/test_pomdp_quantitative_analysis.py
sjunges: I ran the code in the current version and checked those values

volkm: Okay. Then it is maybe more a question for the Storm side (@AlexBork).

PR #263, opened 2025-08-20
Revised configuration files

### LINKED ISSUE #261: Duplicated _config information
We have some duplicated info in https://github.com/moves-rwth/stormpy/blob/master/cmake/config.py.in and https://github.com/moves-rwth/stormpy/blob/master/cmake/info_config.py.in. I wonder which to keep.

volkm:
- Removed a config file which was a remnant of the old CMake build process. Resolves #261
- Moved information on supported libraries to config in info
- Uppercase configuration names


sjunges:
Nice. Do we want to move some of the stuff to info though?  Or do we use the c++ bindings in info for that?

volkm:
Most of the information is already contained in other configurations ([core_config.py.in](https://github.com/moves-rwth/stormpy/blob/master/cmake/core_config.py.in), [pycarl_core_config.py.in](https://github.com/moves-rwth/stormpy/blob/master/cmake/pycarl_core_config.py.in)).
But I can move the information about the installation directories to info as well

sjunges:
Nah, that info is not really relevant anymore. 

What determines which aspects go into core_config and what goes to info_config?

volkm:
There is no clear distinction. We could leave all the pycarl imports and number types in core_config and move the `storm_with_x` to info_config? I think that would be better.

examples/parametric_models/04-parametric-models.py
```diff
@@ -1,19 +1,13 @@
...
-from stormpy import pycarl
-
 import stormpy.examples
 import stormpy.examples.files
-import stormpy._config as config
 
 
 def example_parametric_models_04():
-    # Check support for parameters
-    if not config.storm_with_pars:
```
sjunges, line 12: Is the idea that we now always build with this support?

sjunges: but then, shouldn't we just do import stormpy.pars above?

sjunges:
This looks great! Just one question regarding where we import stormpy.pars

examples/parametric_models/04-parametric-models.py
volkm: No, we can still build without the support. But then the import of stormpy.pars already gives a warning [here](https://github.com/moves-rwth/stormpy/pull/263/files#diff-045ae14b2fee96ae601f07da8eea6970720e8e576d2b72c54f6b5a2f0029ebf8) anyway. This is cleaner and we do not need to manually check the support.

volkm: We should indeed just do the import above. I will change it.

sjunges:
I love this.

Issue #260, opened 2025-08-19
On the compatibility of Storm Python Extensions


sjunges:
This issue is about building projects that extend storm with their own python bindings on top of stormpy. Lets call such a project SP. 
It is not particularly easy to do so and this is also still untested. 

We can consider two cases.
- If stormpy fetches storm, then we have control over what version this is and the source of this version. It would be good if we can get this information from stormpy such that at least we can output a warning if this version is not what SP fetched. Note that Stormpy releases build against a known storm release, and any release of SP will require a specific stormpy release (or set of releases) that uses a fixed storm release.
- If stormpy did not fetch storm but used storm from the system, it is hard to do useful sanity checks, but it would be great if we would know that stormpy was built against a system version of storm and any information about that system version would be great. One can then warn users that if things are not working, that they should consider installing a particular stormpy version (with fetched storm).

I think the whole story is a good argument why building with fetched versions of storm is to be recommended.
Note that if you do pip install stormpy, you always get a fetched version of storm (as a binary) so there we are on the safe side

What this requires:
- [ ] Stormpy should (post installation) record which storm version was used and in particular whether this was a fetched version (then we can be quite sure about its identity) or a system version (then we are in more trouble). 
- [ ] We should clearly map stormpy releases to storm releases for developers of SP.

volkm:
Regarding the last point, I would propose to ensure that Storm and stormpy are always compatible if their major and minor versions are the same. So Storm x.y.z and stormpy x.y.z' are compatible. Then we can use the patch version to introduce independent changes.

sjunges:
We can do that, but that does not resolve the problem that we link in any stormpy version against a concrete storm version and whatever project SP must link against a storm version with the same ABI.

volkm:
That is true, it was more meant for the documentation part.
Would it be sufficient for the linking if we export the information how Storm was installed (wheel/fetch/system) and which Storm commit was used?

sjunges:
I was also talking about the documentation. Let me clarify.

You want to release a version of SP, which will link against some version of storm. Lets call that storm-SP, and say it has version storm x.y.z. 
Now, which  stormpy is compatible with SP? Only any stormpy which is linked against storm with the same ABI as storm-SP.  In particular, a version x.y.z' will almost surely not work, even if stormpy could have also been linked with storm x.y.z'. 
[...]

PR #264, opened 2025-08-20
Storm info

sjunges:
Contributes to https://github.com/moves-rwth/stormpy/issues/260.

(Sorry, appearantly still have to reset my stormpy repo)


lib/stormpy/info/__init__.py
```diff
@@ -12,6 +12,43 @@ def storm_version() -> str:
     return _config.STORM_VERSION
 
 
+def storm_from_system() -> bool:
+    """
+    Does StormPy rely on a (pre-existing) system installation of Storm.
```
volkm:
```suggestion
    Does stormpy rely on a (pre-existing) system installation of Storm.
```

CMakeLists.txt
```diff
@@ -152,11 +193,22 @@ else()
...
 
-
-set(STORM_VERSION ${storm_VERSION})
 # Set number types from Carl
 set_variable_string(STORM_USE_CLN_EA_BOOL ${STORM_USE_CLN_EA})
 set_variable_string(STORM_USE_CLN_RF_BOOL ${STORM_USE_CLN_RF})
+set_variable_string(STORMPY_PRETEND_FETCH ${STORMPY_INFO_PRETEND_FETCH})
+set_variable_string(STORM_DEVELOPER_VERSION ${STORM_VERSION_DEV})
+if (STORM_FETCHED_FROM_REPO)
+    set(STORM_ORIGIN_REPO "\"${STORM_FETCHED_FROM_REPO}\"")
```
volkm, line 202: Do we need quotation marks here? Could we set the quotation marks in info_config.py.in instead (such as for STORM_VERSION)?

volkm (approved):
LGTM. Only two minor questions.

CMakeLists.txt
sjunges: If we do that, we cannot have None there and I wanted the ability to write None as a non-string.

volkm: Right, makes sense. Thanks for the explanation.

PR #266, opened 2025-08-21
Revised CMake

volkm:
- set(STORM_FROM_SYSTEM FALSE) if STORMPY_INFO_PRETEND_FETCH
- prevent issue when Storm is not found in the system


sjunges:
:-D

PR #281, opened 2025-09-25
Revised installation documentation

volkm:
and throw error in CMake if directory hints are given but library is not found.


sjunges:
Great! I did not see any problems in this text.

PR #282, opened 2025-09-26
Revised workflow for creating wheels

### LINKED ISSUE #278: Pypi does not support publishing from reusable workflows
Reusable workflows cannot currently be used as the workflow in a Trusted Publisher for pypi, see [this issue](https://github.com/pypi/warehouse/issues/11096).

The fix would be to integrate the wheelpypi.yml workflow directly into the release workflow.

volkm:
Fixes #278 by moving the upload to pypi into the release workflow.


sjunges:
LGTM!

PR #283, opened 2025-09-26
Build wheels on Linux ARM

volkm:
Support Linux ARM as well.

Requires https://github.com/moves-rwth/carl-storm/pull/94

Seems to work in general, but there are issues with the symbolic engine on ARM, see [this CI test run](https://github.com/volkm/stormpy/actions/runs/18040660533/job/51345290314).

### DIFF
--- a/.github/workflows/create_wheel.yml
+++ b/.github/workflows/create_wheel.yml
@@ -24,7 +24,7 @@ jobs:
     runs-on: ${{ matrix.runs-on }}
     strategy:
       matrix:
-        os: [ linux-intel, macos-intel, macos-arm ]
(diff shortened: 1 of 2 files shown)


sjunges:
I guess we are only setting the threads to one on mac-arm and not on linux arm, yet the arm-issues exist in sylvan 1.8 independent of OS.

sjunges:
I guess https://github.com/moves-rwth/storm/blob/976ebd2ba7eb451b33cb9830294e769445b092c2/src/storm/settings/modules/SylvanSettings.cpp#L51 should be generalised to ARM.

volkm:
Right, that is the point I was missing. I created https://github.com/moves-rwth/storm/issues/788

sjunges:
LGTM!

Issue #251, opened 2025-07-28
Upgrade to pybind version 3.0


volkm:
See the pybind11 [upgrade guide](https://pybind11.readthedocs.io/en/stable/upgrade.html#v3-0) and the [Changelog](https://pybind11.readthedocs.io/en/stable/changelog.html#final-july-10-2025).

Main steps as far as I can see are:

volkm:
It seems, we are already using pybind version 3.0 because we do not fix the pybind version. So far, everything works fine and there are no issues. We might want to follow the steps in the upgrade guide nevertheless.

sjunges:
Do we have to fix this version for releases? @TheGreatfpmK is this something you would be worried about?

volkm:
I think it might be a good idea to fix the pybind version. Then we avoid problems with sudden updates.

TheGreatfpmK:
I'm not sure to be fair. It definitely seems that if we build stormpy and paynt with different pybind we could get some undefined behaviour but I haven't tested that. If you fix the version in stormpy I will do the same in paynt. Alternatively, we could just fix it for releases for now and see if something breaks in the future in the dev branches?

PR #286, opened 2025-10-02
Fixed pybind and scikit-build-core versions

volkm:
Following discussion in https://github.com/moves-rwth/stormpy/issues/251 we fix the versions of pybind and also of scikit-build-core.


sjunges:
The scicit-core-build part is not linked into stormpy, is it? So that one would not need to be fixed? I am fine with it anyhow.

LGTM

volkm:
It is not linked into stormpy, but it could give us more robustness for the build process. But as manually building stormpy will mostly be done by developers, this is not so important.

PR #287, opened 2025-10-02
Use Storm 1.11.1 by default

volkm:
Auto-generated pull request triggered by a new Storm version.
- Manually close and reopen this PR to trigger the CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,7 @@ function(check_hint NAME DIR_FOUND HINT_DIR FOUND_VERSION)
 endfunction(check_hint)
 
 if(ALLOW_STORM_SYSTEM)
-    set(STORM_MIN_VERSION "1.11.1")
+    set(STORM_MIN_VERSION "1.11.0")
     if (ALLOW_STORM_FETCH)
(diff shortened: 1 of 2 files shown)


sjunges:
It shouldn't fail compiling with storm 1.11.0, should it?

CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ function(check_hint NAME DIR_FOUND HINT_DIR FOUND_VERSION)
 endfunction(check_hint)
 
 if(ALLOW_STORM_SYSTEM)
-    set(STORM_MIN_VERSION "1.11.0")
+    set(STORM_MIN_VERSION "1.11.1")
```
sjunges: I don't agree with this change :-)

volkm:
The tests failed because some of the CI test use the Docker `storm:ci` images which are only updated once per day. Waiting until this morning should solve this issue. I was also updating the documentation accordingly.
Another solution could be to explicitly trigger the CI on master after a release.

CMakeLists.txt
volkm: You are right, this is not needed. Let me revert it.

PR #290, opened 2025-10-03
Fix for Intel wheels

volkm:
Needed to bump deploy target for MacOS Intel, see [CI test error](https://github.com/moves-rwth/stormpy/actions/runs/18218167619/job/51871981700).


sjunges:
This also means that we are no longer supporting binaries for some older Mac-os versions with intel-based processors. Not a big thing, maybe worth mentioning in the release notes?

sjunges:
LGTM

volkm:
Yes, I will mention this in the release notes. We could also use the old mac-13 images one last time, but I think that might be overkill.

PR #292, opened 2025-10-06
Fix for STORM_VERSION by taking the STORM_GIT_TAG

volkm:
Use `STORM_GIT_TAG` instead of `STORM_MIN_VERSION`


sjunges:
Thanks for the fix, sorry for not spotting it last time... :-)

volkm:
Sorry for introducing the error in the first place ;)

Issue #301, opened 2025-11-17
Adaptions to changes in storm-pars needed


volkm:
The changes in https://github.com/moves-rwth/storm/pull/640 have changed the storm-pars API and some of the stormpy bindings need to be adapted.

linusheck:
I didn't notice I was assigned, sorry. I'll look into it early next week

volkm:
Thanks for looking into it.

Next time, I will make sure to mention your user name for a notification. Apparently assigning is not enough.

PR #302, opened 2025-11-21
Unified declaration of (parametric) models

volkm:
We had dedicated bindings for parametric sparse and symbolic models.
These changes now use the templated version for better consistency.


sjunges:
LGTM. I have not looked at the failing checks...

volkm:
Failing checks should be due to #301. I will only merge it after the issues were resolved.

PR #314, opened 2025-12-16
CI for weekly wheels

volkm:
The original workflow `create_wheel.yml` needs the Storm version as input argument. This wrapper workflow is scheduled each week to build the wheels for the current master of Storm.

### DIFF
--- a/.github/workflows/create_wheel.yml
+++ b/.github/workflows/create_wheel.yml
@@ -64,7 +64,7 @@ jobs:
           xcode-version: ${{ matrix.xcode }}
       - name: Prepare version
         if: ${{ inputs.dev_version }}
-        run: sed -i -E 's/__version__ = "(.+)"/__version__ = "\1.dev1"/' lib/stormpy/_version.py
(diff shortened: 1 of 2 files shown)


sjunges:
Great!

sjunges:
Should we have something like a version for the nightlies? How do other tools do this?

volkm:
I have seen other people using the timestamps or some hash for the versioning.
But do we need versioning if we do not publish it on pypi?

sjunges:
Not if we trust people to remember that they installed a nightly version :D

volkm:
We could add a [development suffix](https://packaging.python.org/en/latest/specifications/version-specifiers/#developmental-releases) to the stormpy version?

sjunges:
I think that could be good, but it maybe doesnt need to block this pr?

volkm:
The CI now adds a suffix `.dev1` to the stormpy version. An example run can be seen [here](https://github.com/volkm/stormpy/actions/runs/20381321175).

sjunges:
Beautiful! Thanks so much.

PR #315, opened 2025-12-16
Require Python >= 3.10

volkm:
Python 3.9 reached its [end of life](https://devguide.python.org/versions/).


sjunges:
LGTM, but what is this error?

volkm:
Thanks for taking a look at it.

This error occasionally occurred and seems to be somewhat non-deterministic. I informed @AlexBork.

PR #316, opened 2025-12-16
POMDP Test Fix

AlexBork:
The number of cut-off schedulers is non-deterministic on some architectures, therefore we remove the corresponding check in the test cases. This is not a critical aspect to check anyways.


volkm (approved):
LGTM. Thanks for the quick look into it.

PR #322, opened 2026-01-28
Extended support for collecting parameters

volkm:
- Replaced `collect_probability_parameters` by more general `collect_all_parameters`. I think this should be the method to use by default.
- Added binding `collect_rate_parameters`
- Added test for pCTMC

Depends on https://github.com/moves-rwth/storm/pull/858


sjunges:
LGTM. It reminds me that I always wanted to cache the result of this call.

Issue #326, opened 2026-02-05
Use consistent naming for ValueType


volkm:
The stormpy datastructure need different names depending on the underlying value type.
I think the current proposal is to use the following suffixes:
- `` (empty) or `Double`
- `Exact`
- `Parametric`
- `Interval`
We can then use templated functions with the suffix to instantiate all value types.
An example can be seen for the [models](https://github.com/moves-rwth/stormpy/blob/master/src/mod_storage.cpp#L36)

This issue keeps track of the progress to use consistent naming.

PR #327, opened 2026-02-05
Update to use UncertaintyResolutionMode instead of robustUncertainty

volkm:
Adaption to changes in moves-rwth/storm#808

 Uncertainty resolution can be set via `CheckTask.set_uncertainty_resolution_mode` which replaces the previously used `set_robust_uncertainty`.

Moved to templated instantiations for `CheckTask`. The naming is not yet consistent though, see #326.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =============
 
+### Version dev
+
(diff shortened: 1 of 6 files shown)


sjunges:
LGTM, thanks!

PR #328, opened 2026-02-08
Extended support for interval models

volkm:
- Added support for building interval models from Prism files
- Added support for model checking interval DTMCs


lib/stormpy/__init__.py
```diff
@@ -147,7 +216,26 @@ def build_sparse_parametric_model(symbolic_description, properties=None):
...
+
+def build_sparse_interval_model(symbolic_description, properties=None):
+    """
+    Build an interval model in sparse representation from a symbolic description.
+
+    :param symbolic_description: Symbolic model description to translate into a model.
+    :param List[Property] properties: List of properties that should be preserved during the translation. If None, then all properties are preserved.
+    :return: Interval model in sparse representation.
+    """
+    if not symbolic_description.undefined_constants_are_graph_preserving:
```
sjunges, line 230: There is some code duplication here that one may be able to avoid if one uses higher order programming and passes the function (which should be fine?)

sjunges: Alternatively, do we need all the _build_vtype_from or should vtype be an argument?

volkm: I agree that the general structure should be a bit revised. I tried starting this with the valuetype argument but it is only the beginning.
I propose to open a new issue and make a separate PR for this. I think this could be combined with the general restructuring.

sjunges: Agreed. The problem is just becoming more apparent now.

PR #334, opened 2026-02-10
Support for formatting C++ code via clang-format

volkm:
Added support for clang-format for the C++ bindings in the `src` directory. The formatting can be manually triggered with the script `resources/scripts/auto-format.sh`. The script and `.clang-format` are copied from Storm.

Extended the existing formatting CI (for Python) which now also incorporates the C++ formatting.

After merging this PR, we need to initially run the CI for applying the formatting. A quick local check showed that the formatting will lead to some errors, mainly due to the reordering of includes. These will need to be fixed after applying the formatting.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,155 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
+AccessModifierOffset: -1
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveBitFields: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Left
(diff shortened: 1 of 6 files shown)


sjunges:
Do you have an example how it looks like with these formatting?

Also, do we maybe need to order includes semantically and so our formatter should not touch that for the moment?

volkm:
The [formatting PR](https://github.com/volkm/stormpy/pull/6) from my fork.
There will be some include issues which need to be resolved, but this should be doable.
In general, includes can be separated by an empty line into multiple blocks (clang configuration option `IncludeBlocks: Preserve`). Only includes within a block are sorted alphabetically (`SortIncludes: true`). So one can still use empty lines to enforce a certain include order.

sjunges:
LGTM

PR #338, opened 2026-02-21
fix warning and fix support for pybind 3.0.2

### LINKED ISSUE #305: Check that CMake uses FindPython variables

sjunges:
- Fix broken code that no longer compiles with pybind 3.0.2, using documentation from https://pybind11.readthedocs.io/en/stable/advanced/functions.html.
- Fix warning on new cmake regarding findpython, in accordance also with pybind documentation. 

Fixes #305

### DIFF
--- a/src/mod_storage.cpp
+++ b/src/mod_storage.cpp
@@ -22,12 +22,6 @@
 
 PYBIND11_MODULE(_storage, m) {
     m.doc() = "Data structures in Storm";
-
-#ifdef STORMPY_DISABLE_SIGNATURE_DOC
-    py::options options;
-    options.disable_function_signatures();
-#endif
-
     define_bitvector(m);
     define_dd<storm::dd::DdType::Sylvan>(m, "Sylvan");
     define_dd_nt(m);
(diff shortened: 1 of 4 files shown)


tests/storage/test_memorystructure.py
volkm, line 1: Remove this file and add it through the other PR. Otherwise, might get a merge conflict.

src/mod_storage.cpp
volkm, line 1: Is it necessary to remove this part? If yes, we should probably remove it in all other `mod_x` files as well?

src/storage/model.cpp
volkm, line 1: Just for my understanding, what does `cpp_function` change here? Is it needed because we define a property instead of function?

volkm (approved):
LGTM, thanks for taking care of it.
Just three smaller questions.

src/mod_storage.cpp
sjunges: No, that was not necessary (although I think we no longer need that option).

src/storage/model.cpp
sjunges: Indeed. See https://github.com/pybind/pybind11/pull/5533 for explanation on the new pybind static_assert.

volkm: Thanks for the link, that was helpful.

PR #340, opened 2026-02-22
added bindings for memorystructures

sjunges:
Adds bindings for memory structures and an associated test.

### DIFF
--- a/src/storage/memorystructure.cpp
+++ b/src/storage/memorystructure.cpp
@@ -21,7 +21,7 @@ template<typename VT>
 void define_memorystructure_typed(py::module& m, std::string const& vtSuffix) {
     typedef storm::storage::MemoryStructureBuilder<VT> MemoryStructureBuilder;
     py::class_<MemoryStructureBuilder, std::shared_ptr<MemoryStructureBuilder>> msb(m, ("MemoryStructureBuilder" + vtSuffix).c_str());
-    msb.def(py::init<uint_fast64_t, storm::models::sparse::Model<VT> const&, bool>(),  py::arg("nr_memory_states"), py::arg("model"), py::arg("only_initial_states_relevant"));
+    msb.def(py::init<uint_fast64_t, storm::models::sparse::Model<VT> const&, bool>(),  py::arg("nr_memory_states"), py::arg("model"), py::arg("only_initial_states_relevant")=true);
     msb.def("build", &MemoryStructureBuilder::build);
     msb.def("set_label", &MemoryStructureBuilder::setLabel, py::arg("state"), py::arg("label"));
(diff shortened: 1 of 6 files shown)


src/storage/memorystructure.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <storm/storage/memorystructure/SparseModelMemoryProduct.h>
+
+
+void define_memorystructure_untyped(py::module& m) {
+    typedef storm::storage::MemoryStructure MemoryStructure;
+    py::class_<MemoryStructure, std::shared_ptr<MemoryStructure>> memoryStructure(m, "MemoryStructure");
+    memoryStructure.def("product",  [](MemoryStructure& ms, MemoryStructure const& memModel) {return ms.product(memModel);});
+    memoryStructure.def("_product_model_double", [](MemoryStructure& ms, storm::models::sparse::Model<double> const& sparseModel) {return ms.product(sparseModel);});
+    memoryStructure.def("_product_model_exact", [](MemoryStructure& ms, storm::models::sparse::Model<storm::RationalNumber> const& sparseModel) {return ms.product(sparseModel);});
+    memoryStructure.def("_product_model_parametric", [](MemoryStructure& ms, storm::models::sparse::Model<storm::RationalFunction> const& sparseModel) {return ms.product(sparseModel);});
```
volkm, line 15: Interval needed as well?

src/storage/memorystructure.cpp
```diff
@@ -0,0 +1,39 @@
...
+    memoryStructure.def_property_readonly("nr_states", &MemoryStructure::getNumberOfStates);
+    memoryStructure.def_property_readonly("state_labeling", &MemoryStructure::getStateLabeling);
+}
+
+template<typename VT>
+void define_memorystructure_typed(py::module& m, std::string const& vtSuffix) {
+    typedef storm::storage::MemoryStructureBuilder<VT> MemoryStructureBuilder;
+    py::class_<MemoryStructureBuilder, std::shared_ptr<MemoryStructureBuilder>> msb(m, ("MemoryStructureBuilder" + vtSuffix).c_str());
+    msb.def(py::init<uint_fast64_t, storm::models::sparse::Model<VT> const&, bool>(),  py::arg("nr_memory_states"), py::arg("model"), py::arg("only_initial_states_relevant"));
```
volkm: Maybe set a default argument for `only_initial_states_relevant`?

volkm (approved):
LGTM. Just have to see why the test crashes.

src/storage/memorystructure.cpp
sjunges: Memory structures and intervals are currently not supported, but that could delay the ability to do a release.  I'd prefer to add that later (I think the interval code path needs a proper review anyways).

volkm: I was just wondering because there are some interval bindings for the builder. But I agree that it is best to add it later.

volkm:
LGTM.

PR #341, opened 2026-02-22
Fix Python issue on manylinux

volkm:
Need to use `Development.Module` for `find_package(Python ...)`, see the [second warning in the doc](https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode).

Also some tiny cosmetic changes, for example `.idea` occurred twice in the `.gitignore`.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -14,7 +14,6 @@ lib/**/_config.py
 **/__pycache__/
 **/_build/
 **/.pytest_cache/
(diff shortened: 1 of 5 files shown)


sjunges:
LGTM! Thanks, I don't know why i missed this

volkm:
It was only triggered by the wheel CI (good thing we have a weekly check now ;))

PR #342, opened 2026-02-23
Require exact pybind11 version

volkm:
Enforce exact pybind11 version in CMake. This prevents issues with newer pybind versions before supporting them.

Also started to add some documentation on stormpy development.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,7 +48,8 @@ endif()
 # https://pybind11.readthedocs.io/en/latest/compiling.html#findpython-mode
 # Python version should be synced with pyproject.toml
 find_package(Python 3.10 COMPONENTS Interpreter Development.Module REQUIRED)
-find_package(pybind11 CONFIG REQUIRED)
+# Pybind11 version should be synced with pyproject.toml
+find_package(pybind11 3.0.2 EXACT CONFIG REQUIRED)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ endif()
 # https://pybind11.readthedocs.io/en/latest/compiling.html#findpython-mode
 # Python version should be synced with pyproject.toml
 find_package(Python 3.10 COMPONENTS Interpreter Development.Module REQUIRED)
-find_package(pybind11 CONFIG REQUIRED)
+find_package(pybind11 3.0.1 EXACT CONFIG REQUIRED)
```
sjunges: Maybe also add a comment here that the pybind version should be synced with the pyproject.toml?

sjunges:
LGTM! Do we want to have the docs as part of the stormpy website or in a separate folder?

CMakeLists.txt
volkm: Good idea. I have added it.

volkm:
I move the developer documentation to the website folder. I think it would be good to have some general information here, for instance how to add new Python bindings.

volkm:
~I also bumped the pybind version to 3.0.2. I can also do this in a separate PR if it is cleaner.~
Let the [depandabot](https://github.com/moves-rwth/stormpy/pull/345) take care of it.

PR #344, opened 2026-02-23
building models, some refactoring

sjunges:
I started an extended stub for the building models documentation.


volkm:
LGTM. Thanks for starting it.

PR #360, opened 2026-03-09
Statevaluation transform

sjunges:
Matches with https://github.com/moves-rwth/storm/pull/878.


volkm:
LGTM. Can you add a test case for the transformer?

volkm:
The tests running on the ci docker image will still fail because the image is only regenerated once per day. I can manually trigger a build though.

sjunges:
Ah, yes, lets still see whether the fetch works

volkm:
LGTM.  Do you want to add a simple test case or merge it already?

PR #368, opened 2026-03-12
Simplified Binder Dockerfile

volkm:
Use jupyter Docker image as base and use stormpy from pip as defined in `requirements.txt`.


sjunges:
LGTM. Maybe even cleaner to use a requirements.txt :-)

PR #370, opened 2026-03-19
Revised CheckResult and adapted to changes in Storm

volkm:
- use templates for typed CheckResults. We might want to change the naming, but I left it untouched for now.
- adapted to changes in Storm, especially https://github.com/moves-rwth/storm/pull/832/


volkm:
Build currently seems to fail for for macos-14 because Clang instead of AppleClang is used, see [this issue](https://github.com/actions/runner-images/issues/13827).

Seems like this issue might pop up in Storm soon as well.

volkm:
I suggest to merge this for now and open a separate issue for the build issue.

sjunges:
The changes lgtm, thanks for fixing this thoroughly!

PR #371, opened 2026-03-20
CI: revised macOS and Debian configurations

volkm:
- Use macOS 14, 15, 26 and 26-intel, and added `libarchive` dependency
- Added Debian 13


sjunges:
LGTM!

PR #379, opened 2026-04-10
Binding for Storm git hash

volkm:
Extract git hash used for the Storm build


sjunges:
Thanks for your effort and thanks to Filip for testing.

PR #381, opened 2026-04-14
Add two paths to RPath (because stormpy wouldn't import for me)

linusheck:
On my mac, these are the RPaths that are actually correct following the documented installation procedure. Without adding them here, imports fail.


volkm:
Great! Thanks for looking into it, Linus. That error was already annoying me the last days but I missed the fix. So thanks for taking a look into it :)

PR #385, opened 2026-04-21
Add exact interval (RationalInterval) support

lukovdm:
Binds `storm::RationalInterval` throughout: sparse models, matrices, schedulers, model components, model checking, DRN build/export, and the Kwek-Mehlhorn sharpening utility.

Adapts to changes in https://github.com/stormchecker/storm/pull/885


lukovdm:
This PR should be ready for review, it builds on my machine against storm master.

lib/stormpy/__init__.py
```diff
@@ -312,6 +312,26 @@ def build_interval_model_from_drn(file, options=DirectEncodingParserOptions()):
     return _convert_sparse_model(intermediate, value_type=_ValueType.INTERVAL)
 
 
+def build_exact_interval_model_from_drn(file, options=DirectEncodingParserOptions()):
```
volkm, line 351: I think it would be better for maintainability to extend `_convert_sparse_model` to also support exact interval models and then use the standard functions similar to `build_interval_model_from_drn`.

volkm (approved):
Looks good. Only one comment.

And can you add some test cases for exact interval model? Oftentimes the errors only become obvious when actually using the Python bindings.

lib/stormpy/__init__.py
```diff
@@ -103,6 +104,22 @@ def _convert_sparse_model(model, value_type=_ValueType.DOUBLE):
...
+                return model._as_sparse_exact_imdp()
+            elif model.model_type == ModelType.POMDP:
+                return model._as_sparse_exact_ipomdp()
+            elif model.model_type == ModelType.CTMC:
+                return model._as_sparse_exact_ictmc()
+            elif model.model_type == ModelType.MA:
+                return model._as_sparse_exact_ima()
+            elif model.model_type == ModelType.SMG:
+                return model._as_sparse_exact_ismg()
+            else:
```
volkm, line 121: I think these do not exist, right? Better just raise an exception here.

lukovdm:
For the tests, would you rather have me copy existing tests and adapt them for exact intervals or modify the tests to also run for exact intervals?

volkm:
I think we do not have too many tests for intervals anyway at the moment. I suggest to add some tests to storage/test_model and core/test_modelchecking.
I think at a later point, we could think about parameterizing tests for different ValueTypes.

lukovdm:
I have added the tests and indeed found some bugs while testing them.

volkm:
Thanks for adding the tests!
I have manually triggered the Storm build to update the Docker images. Afterwards, I will rerun the remaining CI tests and then merge it. Thanks for your changes.

PR #391, opened 2026-05-06
CI: improved wheel creation

volkm:
- set stormpy version as input
- only build for latest Python version when `dev_version`
- removed scheduled wheel creation (will be handled in separate repo [stormchecker/stormpy-wheels](https://github.com/stormchecker/stormpy-wheels))


sjunges:
Are these Dev versions for the nightlies? I do think that having support for python 3.13 in the nightlies would be good, as not every environment is immediately updated....

volkm:
The dev versions are indeed used for the nightlies. I wanted to limited the amount of builds, and only opted for the latest python version. But I can also add Python 3.13 as well here.

PR #398, opened 2026-05-21
Add STORMPY_prefix to CMake

volkm:
Add `STORMPY_X`prefix to CMake options. Helps with distinguishing from Storm CMake options.

Will maybe require an adaption in PAYNT as well.


sjunges:
Great. I just wondered whether STORMPY_CARL_DIR_HINT or STORMPY_STORM_DIR_HINT is the right way? Should this not be just STORM_DIR_HINT?

volkm:
I agree that it looks a bit weird. I had two main thoughts:
- if everything starts with `STORMPY..` it is easier to see what CMake options are relevant for stormpy
- we avoid a possible name clash with another tool (or Storm itself) which could have a similar option `STORM_DIR_HINT`. For instance, PAYNT could use similar variables.

But I have no strong opinion and could also implement your suggestion.

sjunges:
I think it is a good idea, my main frustration would be if I would need to set STORMPY_STORM_DIR_HINT=... PAYNT_STORM_DIR_HINT=...

volkm:
I can understand it.
@TheGreatfpmK Do you have an opinion from the PAYNT side?

volkm:
After discussion with @TheGreatfpmK we do not use the `STORMPY` prefix for options which are Storm-related (and not stormpy):
- `STORM_DIR_HINT`
- `STORM_GIT_REPO`
- `STORM_GIT_TAG`

sjunges:
LGTM

PR #399, opened 2026-05-21
Replaced #include "storm/.." with #include <storm/..>

### LINKED ISSUE #231: Consistently use "" or <> in includes
Following [this dicsussion](https://github.com/moves-rwth/stormpy/pull/229#discussion_r2146802152) on whether to use `#include ".."` or `#include <..>`.

If I read the [documentation]( https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html) correctly, the quotation marks are only relevant if we want to include header files relative to the current files. This is the case for all stormpy header files. But storm is an external dependency and the files are not relative to the current file. So I would prefer to use angle brackets, but I might be wrong.

Regardless, we should agree on one standard and use it consistently.

volkm:
Fixes https://github.com/stormchecker/stormpy/issues/231

Might need to be adapted if https://github.com/stormchecker/stormpy/pull/367 is merged first.


sjunges:
Nice!  Maybe we should have an automformat rule that there must be an empty line between a block of <> includes and "" includes?  Either way, LGTM!

volkm:
I added a rule in clang-format such that it should now enforce the following order:

```
#include "header.h"

#include <system-library1>
#include <system-library2>

#include "src/core/header1.h"
#include "src/core/header2.h"
```

Automatic sorting can be disabled by
```
// clang-format off
#include <storm/models/sparse/Pomdp.h>  // Must be included before verification.h
// clang-format on

sjunges:
perfect!

PR #401, opened 2026-05-22
Removed unnecessary #ifdef for Python<=2

volkm:
Also removed `src/pycarl/helpers.h` and use `src/helpers.h` instead


sjunges:
LGTM

PR #402, opened 2026-05-26
Exact/Interval EC and Subsystems

sjunges:
Support transformations for subsystems and ECs with exact/interval transitions


volkm (approved):
LGTM.
Just needs formatting and then it can be merged.

PR #403, opened 2026-05-28
Native enum

### LINKED ISSUE #304: Migrate from `py::enum_` to `py::native_enum`

sjunges:
Fixes #304 

For the four enums that had a str from c++,
I went for friendly_name() insstead of name(), as .name is already part of the standard python API.


src/pars/pla.cpp
```diff
@@ -103,14 +103,18 @@ void define_pla(py::module& m) {
         .value("ALLSAT", storm::modelchecker::RegionResult::AllSat)
         .value("ALLVIOLATED", storm::modelchecker::RegionResult::AllViolated)
         .value("UNKNOWN", storm::modelchecker::RegionResult::Unknown)
-        .def("__str__", &streamToString<storm::modelchecker::RegionResult>, py::prepend() /* use custom method instead of default enum overload */);
+        .finalize();
+    m.attr("RegionResult").attr("friendly_name") =
```
volkm, line 107: Just doing something like the following does not work, right?
```
py::native_enum<storm::modelchecker::RegionResult> resultEnum(m, ...
...;
resultEnum.def("friendly_name") ...
```

volkm (approved):
LGTM.

We might want to change all enums to upper case as recommend by the [documentation](https://docs.python.org/3/howto/enum.html#enum-basic-tutorial). But this could be a separate PR.

src/pars/pla.cpp
sjunges: No, sadly not. The native_enum is a builder and the finalize returns void.

PR #408, opened 2026-05-31
Improved CI for creating wheels

volkm:
- added optional inputs `repo` and `ref`. By default the current repo is used, but a custom repo is used for example in the nightly wheel generation in [stormchecker/stormpy-wheels](https://github.com/stormchecker/stormpy-wheels)
- failures are not ignored in development version anymore


sjunges:
LGTM

PR #409, opened 2026-05-31
Fixed wheel generation

volkm:
The wheel generation for the latest release 1.13.1 was not successfull. This PR should fix it.
- added missing dependency libarchive
- also use XCode version 15.4 instead of latest-stable now because the latter lead to compile issues.


sjunges:
LGTM

PR #414, opened 2026-06-10
Consistent use of UPPER_CASE for enums.

### LINKED ISSUE #405: Enum upper case
Change all enums to upper case as recommend by the [documentation](https://docs.python.org/3/howto/enum.html#enum-basic-tutorial).
_Originally posted by @volkm in https://github.com/stormchecker/stormpy/pull/403#pullrequestreview-4379940438_
            

We should probably do this while ensuring "deprecated" old enum names?

volkm:
Fixes https://github.com/stormchecker/stormpy/issues/405.
Not backwards compatible.

I did not add a separate script to check the enum formatting in the CI, but just mention it in the developer's guide.

Contributions from opencode <bot@opencode.ai> (model: opencode/big-pickle).


volkm:
We could think about making a new `dev` branch for the time being which will contain all breaking changes (this PR, restructuring, etc.)
An other argument could be that the `stable` branch is supposed to be stable and not the `master`, so `master` can actually break compatibility.

PR #416, opened 2026-06-11
Replace carl-parser with Lark parser

volkm:
Replaced the C++ bindings of [carl-parser](https://github.com/stormchecker/carl-parser/) with the Python-based [Lark parser](https://github.com/lark-parser/lark). Follows the [suggestion](https://github.com/stormchecker/stormpy/discussions/411) to make it more maintainable.

- lark is an optional dependency and can be installed with `pip install stormpy[parser]`.
- the lark grammar is tightly aligned with the carl-parser [grammar](https://github.com/stormchecker/carl-parser/blob/master/src/carl-parser/Serialization.g4). A notable difference is that I removed support for `.` in variable names. Otherwise `- 2.5` could be parsed as the subtraction of `2` and a variable named `.5`. The variable names currently still support a long list of symbols which I find unnecessary complicated, but I opted to follow the original carl-parser grammar.
- arithmetic operations first try to apply the operation on the given types, afterwards try the reverse order, and lastly cast everything to `Polynomial` and try again. This is relatively simple to implement, but could be too expensive in certain cases.

Most implementation work was done by opencode [bot@opencode.ai](mailto:bot@opencode.ai) (model: opencode/big-pickle).


pyproject.toml
```diff
@@ -34,11 +34,18 @@ Documentation = "https://stormchecker.github.io/stormpy"
...
 ]
+parser = [
+    "lark"
+]
 plot = [
     "matplotlib",
     "numpy",
     "scipy",
 ]
+dev = [
```
sjunges, line 45: is this related?

.github/workflows/Dockerfile.fetch
sjunges, line 1: i dont understand what this did and why we no longer need it :-)

sjunges:
Great progress! LGTM

pyproject.toml
volkm: Not too much. I just changed the order to have `dev` come before `doc`

.github/workflows/Dockerfile.fetch
volkm: It was not really different from the main Dockerfile. There might have been changes at an earlier point, was introduced in https://github.com/stormchecker/stormpy/pull/229

PR #420, opened 2026-06-29
Revised CI for wheel creation

volkm:
- Fixed cmake.args. In the old version, setting the build type overwrote all other options.
- Build wheels with multiple threads governed by NR_JOBS

Minor:
- Print CMAKE_BUILD_TYPE
- Removed quotation marks for cmake.args
- Use same ordering of cmake options in CMakeLists.txt and pyproject.yml

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,16 +84,17 @@ sdist.include = [
 STORM_DIR_HINT=""
 ALLOW_STORM_SYSTEM="ON"
 ALLOW_STORM_FETCH="ON"
+STORM_GIT_REPO="https://github.com/stormchecker/storm.git"
+# Storm version should be synced with STORM_MIN_VERSION in CMakeLists.txt
+STORM_GIT_TAG="1.13.0"
 USE_STORM_DFT="ON"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -84,16 +84,17 @@ sdist.include = [
...
 USE_STORM_PARS="ON"
 USE_STORM_POMDP="ON"
 USE_CLN_NUMBERS="ON"
-COMPILE_WITH_CCACHE="ON"
 STORMPY_DISABLE_SIGNATURE_DOC="OFF"
-STORM_GIT_REPO="https://github.com/stormchecker/storm.git"
-# Storm version should be synced with STORM_MIN_VERSION in CMakeLists.txt
-STORM_GIT_TAG="1.13.0"
+COMPILE_WITH_CCACHE="ON"
+STORMPY_INFO_PRETEND_FETCH="OFF"
```
sjunges, line 99: This may profit from an explanation what this variable does

volkm: Added a comment.

PR #424, opened 2026-07-25
Feature/valuations

sjunges:
Match https://github.com/stormchecker/storm/pull/937, requires https://github.com/stormchecker/storm/pull/949 also to compile.


sjunges:
@volkm this wont compile unless I start merging the other PRs in Storm, but if you could have a look whether there is something major that you want changed, then we could avoid having the main thing nto compiling for too long?

src/pomdp/transformations.cpp
```diff
@@ -75,9 +75,9 @@ template<typename ValueType>
 void define_transformations_int(py::module &m, std::string const &vtSuffix) {
     py::class_<storm::pomdp::ObservationTraceUnfolder<ValueType>> unfolder(m, ("ObservationTraceUnfolder" + vtSuffix).c_str(),
                                                                            "Unfolds observation traces in models");
-    unfolder.def(py::init<storm::models::sparse::Pomdp<ValueType>, std::vector<ValueType> const &, std::shared_ptr<storm::expressions::ExpressionManager> &,
+    unfolder.def(py::init<storm::models::sparse::Pomdp<ValueType>, std::vector<ValueType> const &, std::shared_ptr<storm::expressions::ExpressionManager>,
```
volkm, line 78: What is the reason for changing it to call-by-copy?

tests/storage/test_valuations.py
```diff
@@ -3,24 +3,159 @@
 import pytest
 
 
-class TestStateValuationTransformer:
+class TestValuationTransformer:
```
volkm, line 6: Also rename the test file?

tests/storage/test_valuations.py
```diff
@@ -3,24 +3,159 @@
...
+        # mismatch). It must now raise a catchable exception instead.
+        manager = stormpy.ExpressionManager()
+        var_r = manager.create_rational_variable(name="r")
+        builder = stormpy.ValuationDescriptionBuilder(manager)
+        builder.add_double_variable(var_r)
+        desc = builder.build_class_description()
+        vals = stormpy.Valuations(desc, manager, 1)
+        vals.write_double_value(0, var_r, 0.5)
+
+        with pytest.raises(RuntimeError):
```
volkm, line 121: I am not sure I understand why it raises an exception here. If the type mismatches, should there not already be an error when adding a double variable of type rational variable?

lib/stormpy/storage/__init__.py
```diff
@@ -114,47 +230,151 @@ def _get_boolean_value(self, state, var):
...
 @deprecated("Use general method 'get_values_states' instead.", version="1.10.0")
 def _get_rational_values_states(self, var):
     return self._get_rational_values_states(var)
 
 
-StateValuation.get_rational_values_states = _get_rational_values_states
+Valuations.get_rational_values_states = _get_rational_values_states
+
+
+@deprecated("Use get_nr_of_entities instead.", version="1.10.0")
```
volkm:
```suggestion
@deprecated("Use get_nr_of_entities instead.", version="1.13.2")
```
Indicates the version where this was deprecated.

lib/stormpy/storage/__init__.py
```diff
@@ -114,47 +230,151 @@ def _get_boolean_value(self, state, var):
...
+    return self.get_nr_of_entities()
+
+
+Valuations.get_nr_of_states = _get_nr_of_states
+
+# Deprecated class alias
+StateValuation = Valuations
+
+
+@deprecated("Use ValuationDescriptionBuilder and Valuations instead.", version="1.10.0")
```
volkm:
```suggestion
@deprecated("Use ValuationDescriptionBuilder and Valuations instead.", version="1.13.2")
```

lib/stormpy/storage/__init__.py
```diff
@@ -114,47 +230,151 @@ def _get_boolean_value(self, state, var):
...
+
+        num_entities = max(self._states.keys()) + 1 if self._states else 0
+        valuations = Valuations(class_description, self._manager, num_entities)
+        for state, values in self._states.items():
+            for var, val in values.items():
+                valuations.write_value(state, var, val)
+        return valuations
+
+
+@deprecated("Use ValuationTransformer instead.", version="1.10.0")
```
volkm:
```suggestion
@deprecated("Use ValuationTransformer instead.", version="1.13.2")
```

lib/stormpy/storage/__init__.py
```diff
@@ -114,47 +230,151 @@ def _get_boolean_value(self, state, var):
...
+class StateValuationTransformer(ValuationTransformer):
+    """
+    Deprecated shim for the old state-valuation transformer API.
+    """
+
+    @deprecated("Use general method 'add_expression' instead.", version="1.10.0")
```
volkm:
```suggestion
    @deprecated("Use general method 'add_expression' instead.", version="1.13.2")
```

lib/stormpy/storage/__init__.py
```diff
@@ -114,47 +230,151 @@ def _get_boolean_value(self, state, var):
...
+    def add_boolean_expression(self, expression_variable, defining_expression):
+        self.add_expression(expression_variable, defining_expression)
+
+    @deprecated("Use general method 'add_expression' instead.", version="1.10.0")
```
volkm:
```suggestion
    @deprecated("Use general method 'add_expression' instead.", version="1.13.2")
```

src/storage/valuation.cpp
```diff
@@ -2,63 +2,175 @@
...
-             py::arg("rational_values") = std::vector<storm::RationalNumber>(), "Adds a new state, no more variables should be added afterwards")
-        .def("build", &storm::storage::sparse::StateValuationsBuilder::build, "Creates the finalized state valuations object");
+    py::class_<Valuations, std::shared_ptr<Valuations>>(m, "Valuations", "Valuations for explicit entities (states/observations)")
+        .def(py::init([](ValuationClassDescription const& desc, storm::expressions::ExpressionManager& manager, uint64_t numEntities) {
+                 return std::make_shared<Valuations>(desc, manager.getSharedPointer(), numEntities);
+             }),
+             py::arg("class_description"), py::arg("manager"), py::arg("num_entities") = uint64_t(0),
+             "Construct Valuations from a class description, allocating storage for num_entities entities")
+        .def("_get_boolean_value", &Valuations::getBooleanValue, py::arg("entity"), py::arg("variable"))
+        .def("_get_int64_value", &Valuations::getInt64Value, py::arg("entity"), py::arg("variable"))
```
volkm, line 117: Does this throw an error when the number does not fit in int64?

src/storage/valuation.cpp
```diff
@@ -2,63 +2,175 @@
...
+                // Valuations::getBooleanValues returns a BitVector, not a std::vector<bool>, so it
+                // doesn't get pybind11's automatic std::vector<T> -> list conversion (unlike the
+                // int64/rational siblings below). We can't return the BitVector as-is either: unlike a
+                // dense per-entity list, iterating a BitVector yields the *indices* of its true entries
+                // (see storm::storage::BitVector's documented iterator semantics), which would silently
+                // change the meaning of get_values_states() for booleans compared to the other two types.
+                // This copy is therefore O(entities) on every call: cheap for typical models, but avoid
+                // calling this in a hot loop over many boolean variables for models with millions of
+                // states/entities -- cache the result instead, or use
+                // _get_boolean_values_states_as_bitvector if a BitVector is acceptable.
```
volkm: I think this should be part of the documentation.

src/storage/valuation.cpp
```diff
@@ -2,63 +2,175 @@
...
+                storm::storage::BitVector bv = v.getBooleanValues(var);
+                std::vector<bool> result(bv.size(), false);
+                for (uint64_t index : bv) {
+                    result[index] = true;
+                }
+                return result;
```
volkm: Should this maybe be part of Storm to convert from/to `std::vector<bool>`?

src/storage/valuation.cpp
```diff
@@ -2,63 +2,175 @@
...
+        .def("get_all_variables", &Valuations::getAllVariables, "Get all variables with valuations")
+        .def("entity_has_variable", &Valuations::entityHasVariable, py::arg("entity"), py::arg("variable"))
+        .def("select_entities", py::overload_cast<storm::storage::BitVector const&>(&Valuations::selectEntities, py::const_), py::arg("selected_entities"))
+        .def("select_entities", py::overload_cast<std::vector<uint64_t> const&>(&Valuations::selectEntities, py::const_), py::arg("selected_entities"))
+        .def(
+            "write_bool_value",
+            [](Valuations& v, uint64_t entity, storm::expressions::Variable const& var, bool value) {
+                if (entity >= v.getNumberOfEntities())
+                    throw py::index_error();
+                v.getStorage().writeValue<bool>(entity, var, value);
```
volkm, line 189: Is Storm throwing an error it the types do not match?

volkm (changes requested):
Overall looks good to me. Mostly some smaller comments.

On a higher level, I was wondering whether it is needed to define some bindings in the C++ part and then add some extension in `__init__.py`? Would it be possible to define the binding only in the C++ part and add the type check there as well? This would make it easier if the code is not spread over two locations.

src/pomdp/transformations.cpp
sjunges: Memory issues on the C++ side, this is reflecting a change. The deeper issue is with where our expressions live, we have a very old bugreport regarding this in storm

tests/storage/test_valuations.py
sjunges: Yes.

src/storage/valuation.cpp
sjunges: Pretty sure it does.

src/storage/valuation.cpp
sjunges: For the C++ api, I dont see an advantage in doing that, but for the python api (where types are hidden) I think this is important to avoid bugs.

src/storage/valuation.cpp
sjunges: I will check.

src/pomdp/transformations.cpp
volkm: Thanks, that is the change from https://github.com/stormchecker/storm/pull/949

src/storage/valuation.cpp
volkm: Sorry, I meant whether we should add a dedicated method `BitVector::toVector` in Storm which returns a `std::vector<bool>`. This method could then be used in this binding.
The idea would be that such a dedicated `toVector` method could also be useful in other places.

src/storage/valuation.cpp
sjunges: it does.

src/storage/valuation.cpp
sjunges: I havent done that yet.  Can do that :-)

tests/storage/test_valuations.py
volkm: Have a better understanding of the code now and it makes sense. It is a rational variable but the value is set as a double (which is allowed). But then reading it again as a rational number raises the error.

lib/stormpy/storage/__init__.py
```diff
@@ -70,42 +70,10 @@ def get_maximal_end_components(model):
...
-    else:
-        raise ValueError(f"Variable {var} has unsupported type")
-
-
-StateValuation.get_value = _get_value
-StateValuation.get_values_states = _get_values_states
+# Extend class Valuations
+# get_value, get_values_states, write_value, get_boolean_values_states_as_bitvector,
+# get_integer_bigint_value/write_integer_bigint_value, and get_double_value/get_double_values_states/
+# write_double_value (including their type checks) are implemented directly in src/storage/valuation.cpp.
```
volkm, line 76: Remove?

volkm (approved):
Thanks for the changes, in particular moving the Python code to the C++ side.
Looks good to me now.

volkm:
CI seems to improve, only one example is failing. Seems good to have additional checks now.

PR #437, opened 2026-08-18
Towards generating good stubs: attempt at fixing C++ name leaks by first declaring, then defining

linusheck:
Currently, the pybind11 leaks C++ type names, which makes it very difficult to generate stubs. This PR fixes this, but I'm not sure if this is the best way to do it yet. We first declare everything and then define the methods. Because the core, storage, logic, and utility modules all reference each other, this PR creates a new `mod_bindings.cpp` that technically is a joint module containing all these modules (the visible module paths themselves stay the same). All of the extension modules are fixed to do declare-then-define as well, but these modules don't reference each other, so that has just been done in the module directly. A couple of individual fixes were needed for definitions that were missing.

On top of this PR, it should be quite straightforward to generate stubs :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -332,14 +332,28 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/pycarl/definitions.h.in ${CMAKE_C
 
 # Build stormpy
 ######
-# Build stormpy modules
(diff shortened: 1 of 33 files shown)


linusheck:
I'll try to generate stubs on top of this PR.

linusheck:
We'll revisit this after we are happy with stormpy's module structure.

PR #441, opened 2026-08-20
Better Sphinx docs: example improvement on logic

linusheck:
This PR contains improvements to the Sphinx docs. There are a few global improvements:

- Moving the imports around in `__init__.py` removes most C++ class names
- Some improvements in `conf.py`

As an example on how to restructure the docs, I added an overview page and separate pages for the logic module:
image

The subclasses now show inherited properties and methods on their own pages:
image


doc/source/conf.py
```diff
@@ -33,6 +36,11 @@
 # Autodoc options
 autoclass_content = "both"  # Add documentation for both the class and __init__
 
+# Display e.g. "BitVector" instead of "stormpy.storage.BitVector"
+python_use_unqualified_type_names = True
```
volkm: Is the module name still available in the main description of the corresponding class? Otherwise it becomes confusing to which module a class belongs.

doc/source/conf.py
```diff
@@ -138,3 +146,16 @@
 myst_enable_extensions = [
     "colon_fence",
 ]
+
+# The following code makes Sphinx display e.g. "Environment()" instead of "<stormpy.Environment object at 0x10abc123>"
```
volkm: Am I right that the `X object at Y` only occurs when using a class as default argument? [Mutable default arguments can lead to issues](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) and we should probably rewrite it anyway.

tests/test_api_documentation.py
volkm, line 1: I am not sure I understand what this test does. Does it check that all classes are also appearing in the documentation?

In general, this seems only relevant for the documentation. In particular, what happens if running the tests but without generating any documentation?
Maybe we should move it to `resources/scripts` or in `doc`.

doc/source/api/logic.rst
volkm, line 1: I think it is nice to have such a structured overview of the classes. I am only a bit worried because classes now have to be manually specified and need to be maintained.

.gitignore
```diff
@@ -14,3 +14,4 @@ cmake-build-debug/
 .cache/
 .DS_Store
 .venv/
+doc/source/api/generated/
```
volkm, line 17: Is this where the api doc is generated? I thought it would be in a separate `build` directory as before.

doc/source/api/logic.rst
linusheck: Indeed, one needs to keep a manual list of every class in every module. I've added a test checking that there is an entry for every class, so that test should fail if we are missing a class. Alternatively, I could just add an "everything else" list

tests/test_api_documentation.py
linusheck: Indeed. Agreed, this should be moved :D

doc/source/conf.py
linusheck: Indeed, I made an issue for that: https://github.com/stormchecker/stormpy/issues/443
and will remove this.

.gitignore
linusheck: This is where the `rst` files generated by `toctree` are, this is different from the generated HTML which is in `build`. So this is the conventional way to do this.

doc/source/conf.py
volkm: Thanks for opening the issue

.gitignore
sjunges: should we call it api/.generated/ then?

linusheck: Probably not, because this is the public link to the page (e.g. api/generated/stormpy.logic.BinaryPathFormula.html)

doc/source/conf.py
linusheck: indeed, removed this

PR #442, opened 2026-08-20
Migrate `unique_ptr` and `shared_ptr` uses to `py::smart_holder`

### LINKED ISSUE #303: Use `py::smart_holder`

linusheck:
Fixes #303. I ran this script on the source code:

```
#!/usr/bin/env python3
import re
from pathlib import Path


def split_args(s):
    parts, start, depth = [], 0, 0
    for i, c in enumerate(s):
        depth += c in "<([{"
        depth -= c in ">)]}"
        if c == "," and depth == 0:
            parts.append(s[start:i])
            start = i + 1
    return parts + [s[start:]]


for path in Path("src").rglob("*"):
    if path.suffix not in {".cpp", ".h"}:
        continue
    text, pos = path.read_text(), 0
    while (start := text.find("py::class_<", pos)) >= 0:
        if text[text.rfind("\n", 0, start) + 1 : start].lstrip().startswith("//"):
            pos = start + 1
            continue
        end, depth = start + len("py::class_"), 0
        while end < len(text):
            depth += text[end] == "<"
            depth -= text[end] == ">"
            if depth == 0:
                break
            end += 1
        args = split_args(text[start + len("py::class_<") : end])
        args = [a for a in args if not re.fullmatch(r"\s*std::(?:shared|unique)_ptr<.*>\s*", a, re.S)]
        replacement = "py::classh<" + ",".join(args) + ">"
        text = text[:start] + replacement + text[end + 1 :]
        pos = start + len(replacement)
    text = re.sub(r"(?m)^PYBIND11_DECLARE_HOLDER_TYPE\(T, std::shared_ptr<T(?: const)?>\)\n\n?", "", text)
    path.write_text(text)
```

### DIFF
--- a/src/core/multiobjective.cpp
+++ b/src/core/multiobjective.cpp
@@ -27,8 +27,7 @@ void define_multiobjective(py::module& m, std::string const& vtSuffix) {
           py::arg("model"), py::arg("formula"), py::arg("compute_scheduler") = false);
 
     using PcaaWeightVectorChecker = storm::modelchecker::multiobjective::PcaaWeightVectorChecker<storm::models::sparse::Mdp<ValueType>>;
-    py::class_<PcaaWeightVectorChecker, std::unique_ptr<PcaaWeightVectorChecker>> weightedObjectiveMdpModelChecker(
-        m, ("WeightedObjectiveMdpModelChecker" + vtSuffix).c_str());
+    py::classh<PcaaWeightVectorChecker> weightedObjectiveMdpModelChecker(m, ("WeightedObjectiveMdpModelChecker" + vtSuffix).c_str());
     weightedObjectiveMdpModelChecker.def("check", &PcaaWeightVectorChecker::check, py::arg("env"), py::arg("weight_vector"))
         .def("get_achievable_point", &PcaaWeightVectorChecker::getAchievablePoint)
         .def("get_optimal_weighted_sum", &PcaaWeightVectorChecker::getOptimalWeightedSum,
--- a/src/common.h
+++ b/src/common.h
@@ -11,7 +11,4 @@
 namespace py = pybind11;
 using namespace pybind11::literals;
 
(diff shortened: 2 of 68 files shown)


volkm:
[Documentation](https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html)

volkm (approved):
LGTM.

volkm:
@sjunges Any comments or things to consider before merging?

src/core/multiobjective.cpp
```diff
@@ -27,8 +27,7 @@ void define_multiobjective(py::module& m, std::string const& vtSuffix) {
           py::arg("model"), py::arg("formula"), py::arg("compute_scheduler") = false);
 
     using PcaaWeightVectorChecker = storm::modelchecker::multiobjective::PcaaWeightVectorChecker<storm::models::sparse::Mdp<ValueType>>;
-    py::class_<PcaaWeightVectorChecker, std::unique_ptr<PcaaWeightVectorChecker>> weightedObjectiveMdpModelChecker(
```
volkm, line 30: This seems to be the only case where we used a `unique_ptr`instead of a `shared_ptr`. I would argue that using the smart ~pointer~ holder here is still a good idea.

linusheck: With the smart pointer, you mean the `unique_ptr` or the `smart_holder`? :)

volkm: Smart holder :)

sjunges: Why not use the smart holder everywhere? I think the main problem could be on places where we have problems with the ownership?

sjunges:
LGTM

src/core/multiobjective.cpp
volkm: We are now actually using the smart holder everywhere. I just wanted to point out that we originally used a `unique_ptr` here, in case this was relevant. But I believe it was not important.

PR #448, opened 2026-08-30
Use jupytext and replace notebooks by Markdown

volkm:
The documentation now uses [Myst Markdown](https://mystmd.org/) together with [Jupytext](https://jupytext.org/) to support Jupyter notebooks. This allows better git integration.
The existing notebooks have been removed, but the command `make notebooks` automatically builds all notebooks next to their corresponding Markdown files.
Use `make check-doc` to ensure that all generated notebooks can be executed successfully. This is also checked in the CI.
Use `make format` to execute formatting: for cpp files via the script in resources, for Python via black and for the Markdown documentation via black as well.


linusheck:
LGTM!

PR #449, opened 2026-09-01
Consistent and readable generics

linusheck:
Adds a new generic handling that automatically generates instantiations for the given types and then puts a wrapper (in _template.py) around it to allow for this functionality:

```
# Readable aliases:
assert stormpy.dft._dft._DFT_Double is stormpy.dft.DFT[float]
assert stormpy.dft._dft._DFT_RationalFunction is stormpy.dft.DFT[stormpy.RationalFunction]

path = stormpy.examples.files.dft_json_and
double_dft = stormpy.dft.load_dft_json_file(path)
assert type(double_dft) is stormpy.dft.DFT[float]

rational_function_dft = stormpy.dft.load_parametric_dft_json_file(path)
assert type(rational_function_dft) is stormpy.dft.DFT[stormpy.RationalFunction]

# Construct a specialization explicitly:
explicit_builder = stormpy.dft.ExplicitDFTModelBuilder[float](double_dft)
assert type(explicit_builder) is stormpy.dft.ExplicitDFTModelBuilder[float]

# Or deduce the specialization from the constructor argument:
inferred_builder = stormpy.dft.ExplicitDFTModelBuilder(rational_function_dft)
assert type(inferred_builder) is stormpy.dft.ExplicitDFTModelBuilder[stormpy.RationalFunction]

# Overloaded functions call the specialization:
concrete_model = stormpy.dft.build_model(double_dft)
assert not concrete_model.supports_parameters

parametric_model = stormpy.dft.build_model(rational_function_dft)
assert parametric_model.supports_parameters
```

This replaces the custom Python wrapper functions that currently check input types.

This is a breaking API change for version 2.0. Let's discuss whether to re-export the old type names.

This PR migrates `pars` and `dft`, more PRs that I will open on top of this will migrate the rest of stormpy.


volkm:
Thanks for adding these changes. Can you maybe rebase from the latest master? Then it is easier to see your changes.

tests/dft/test_transformations.py
```diff
@@ -14,7 +14,10 @@ def test_instantiate_dft(self):
         assert dft.nr_elements() == 7
         assert dft.nr_be() == 4
 
+        instantiator_type = stormpy.dft.DFTInstantiator[stormpy.RationalFunction, float]
         instantiator = stormpy.dft.DFTInstantiator(dft)
+        assert type(instantiator) is instantiator_type
+        assert instantiator_type is stormpy.dft._dft._DFTInstantiator_RationalFunction_Double
```
volkm, line 20:
```suggestion
        assert instantiator_type is stormpy.dft._DFTInstantiator_RationalFunction_Double
```
Would this also work?

tests/dft/test_dft.py
```diff
@@ -9,6 +9,29 @@
 
 @dft
 class TestDft:
+    def test_generic_dft_type(self):
+        dft = stormpy.dft.load_dft_json_file(get_example_path("dft", "and.json"))
+
+        concrete_type = stormpy.dft.DFT[float]
+        assert type(dft) is concrete_type
+        assert concrete_type is stormpy.dft._dft._DFT_Double
```
volkm, line 17:
```suggestion
        assert concrete_type is stormpy.dft._DFT_Double
```
Would this also work?

tests/dft/test_io.py
```diff
@@ -60,3 +60,28 @@ def test_export_dft_json_file(self, tmpdir):
...
+    def test_export_parametric_dft_json_string(self):
+        dft = stormpy.dft.load_parametric_dft_galileo_file(get_example_path("dft", "symmetry_param.dft"))
+        parameter_names = {parameter.name for parameter in stormpy.dft.get_parameters(dft)}
+        assert parameter_names == {"x", "y"}
+
+        json_string = stormpy.dft.export_dft_json_string(dft)
+        dft2 = stormpy.dft.load_parametric_dft_json_string(json_string)
+
+        assert type(dft2) is stormpy.dft.DFT[stormpy.RationalFunction]
+        assert dft2.nr_elements() == dft.nr_elements()
```
volkm, line 73:
```suggestion
        assert dft2.nr_elements() == dft.nr_elements()
        assert dft2.nr_be() == dft.nr_be()
        assert dft2.nr_dynamic() == dft.nr_dynamic()
```

tests/dft/test_io.py
```diff
@@ -60,3 +60,28 @@ def test_export_dft_json_file(self, tmpdir):
...
+        dft = stormpy.dft.load_parametric_dft_galileo_file(get_example_path("dft", "symmetry_param.dft"))
+        parameter_names = {parameter.name for parameter in stormpy.dft.get_parameters(dft)}
+        assert parameter_names == {"x", "y"}
+
+        export_file = os.path.join(str(tmpdir), "symmetry_param.json")
+        stormpy.dft.export_dft_json_file(dft, export_file)
+        dft2 = stormpy.dft.load_parametric_dft_json_file(export_file)
+
+        assert type(dft2) is stormpy.dft.DFT[stormpy.RationalFunction]
+        assert dft2.nr_elements() == dft.nr_elements()
```
volkm, line 88:
```suggestion
        assert dft2.nr_elements() == dft.nr_elements()
        assert dft2.nr_be() == dft.nr_be()
        assert dft2.nr_dynamic() == dft.nr_dynamic()
```

examples/dfts/03-generic-types.py
```diff
@@ -0,0 +1,37 @@
...
+
+import stormpy
+import stormpy.dft
+import stormpy.examples.files
+
+
+def demonstrate_generic_dft_types():
+    # Readable aliases:
+    assert stormpy.dft._dft._DFT_Double is stormpy.dft.DFT[float]
+    assert stormpy.dft._dft._DFT_RationalFunction is stormpy.dft.DFT[stormpy.RationalFunction]
```
volkm, line 11:
```suggestion
    assert stormpy.dft._DFT_Double is stormpy.dft.DFT[float]
    assert stormpy.dft._DFT_RationalFunction is stormpy.dft.DFT[stormpy.RationalFunction]
```

tests/dft/test_dft.py
```diff
@@ -9,6 +9,29 @@
...
+
+        explicit_copy = stormpy.dft.DFT[float](dft)
+        inferred_copy = stormpy.dft.DFT(dft)
+        keyword_inferred_copy = stormpy.dft.DFT(dft=dft)
+        assert type(explicit_copy) is concrete_type
+        assert type(inferred_copy) is concrete_type
+        assert type(keyword_inferred_copy) is concrete_type
+
+        builder = stormpy.dft.ExplicitDFTModelBuilder(dft)
+        assert type(builder) is stormpy.dft.ExplicitDFTModelBuilder[float]
```
volkm, line 33: Can we also add a test for invalid inputs such as `explicit_builder = stormpy.dft.ExplicitDFTModelBuilder[storm.Rational](double_dft)`? I guess this should throw some exception?

src/binding_type_index.h
```diff
@@ -0,0 +1,81 @@
...
+struct BindingTypeArgument<storm::RationalNumber> {
+    static TemplateArgument get() {
+        return {pybind11::module_::import("stormpy").attr("Rational"), "Rational"};
+    }
+};
+
+template<>
+struct BindingTypeArgument<storm::RationalFunction> {
+    static TemplateArgument get() {
+        return {pybind11::module_::import("stormpy").attr("RationalFunction"), "RationalFunction"};
```
volkm, line 31: This revealed an [important issue](https://github.com/stormchecker/stormpy/issues/450) in stormpy: `storm::RationalFunction` is mapped to `stormpy.FactorizedRationalFunction` and not to `stormpy.RationalFunction`. This will be fixed by https://github.com/stormchecker/stormpy/pull/451 such that this code should be correct without any change.

src/binding_type_index.h
```diff
@@ -0,0 +1,81 @@
...
+};
+
+template<typename... ValueTypes>
+TemplateIndex typeIndex() {
+    return makeTemplateIndex({bindingTypeArgument<ValueTypes>()...});
+}
+
+template<storm::models::ModelType ModelKind, typename... ValueTypes>
+TemplateIndex typeIndex() {
+    return makeTemplateIndex({BindingModelTypeArgument<ModelKind>::get(), bindingTypeArgument<ValueTypes>()...});
```
volkm: Is `bindingTypeArgument<ValueTypes>()` needed or could it be replaced by `BindingTypeArgument<ValueTypes>::get()`? Then we could remove `bindingTypeArgument` which only differs from `BindingTypeArgument` by lower case.

src/binding_type_index.h
```diff
@@ -0,0 +1,81 @@
...
+#include <storm/adapters/RationalFunctionAdapter.h>
+#include <storm/adapters/RationalNumberAdapter.h>
+#include <storm/models/ModelType.h>
+
+#include "src/template_binding.h"
+
+namespace stormpy::bindings {
+
+template<typename ValueType>
+struct BindingTypeArgument;
```
volkm, line 12: Maybe call it `BindingValueTypeArgument`? There are more types: model type, reward type, ...

src/template_binding.h
```diff
@@ -0,0 +1,165 @@
...
+ * @tparam Extra Types of additional constructor arguments forwarded to
+ *     `pybind11::class_`, such as base-class handles.
+ * @param module Native module receiving the class.
+ * @param name Complete native Python class name.
+ * @param description Python class docstring.
+ * @param extra Additional arguments forwarded to `pybind11::class_`.
+ * @return The class binding, ready for `.def(...)` calls.
+ */
+template<typename Class, typename... Options, typename... Extra>
+pybind11::class_<Class, Options...> bindInternalClass(pybind11::module_& module, std::string const& name, char const* description, Extra&&... extra) {
```
volkm: Should it not be `classh`?

volkm: And could `char const*` be replaced by `std::string const&` as well? Or by `std::string_view`?

src/template_binding.h
```diff
@@ -0,0 +1,165 @@
...
+ * @param extra Additional arguments forwarded to `pybind11::class_`.
+ * @return The class binding, ready for `.def(...)` calls.
+ * @throws std::runtime_error If the family already contains `index.key`.
+ */
+template<typename Class, typename... Options, typename... Extra>
+pybind11::class_<Class, Options...> bindTemplateClass(pybind11::module_& module, std::string_view family, TemplateIndex const& index, char const* description,
+                                                      Extra&&... extra) {
+    pybind11::dict instantiations = templateInstantiations(module, family);
+    if (instantiations.contains(index.key)) {
+        throw std::runtime_error("Duplicate native template instantiation for " + std::string(family));
```
volkm:
```suggestion
        throw std::runtime_error("Duplicate native template instantiation for " + std::string(family) + " and " + index.label);
```

src/dft/analysis.cpp
```diff
@@ -61,24 +65,25 @@ void define_analysis_typed(py::module& m, std::string const& vt_suffix) {
...
 
-    m.def(("_analyze_dft" + vt_suffix).c_str(), &analyzeDFT<ValueType>, "Analyze the DFT", py::arg("dft"), py::arg("properties"), py::arg("symred") = true,
+    m.def("analyze_dft", &analyzeDFT<ValueType>, "Analyze the DFT", py::arg("dft"), py::arg("properties"), py::arg("symred") = true,
           py::arg("allow_modularisation") = false, py::arg("relevant_events") = storm::dft::utility::RelevantEvents(),
           py::arg("allow_dc_for_relevant") = false);
 
-    m.def(("_build_model" + vt_suffix).c_str(), &buildModel<ValueType>, "Build state-space model (CTMC or MA) for DFT", py::arg("dft"), py::arg("symmetries"),
-          py::arg("relevant_events") = storm::dft::utility::RelevantEvents(), py::arg("allow_dc_for_relevant") = false);
+    m.def("build_model", &buildModel<ValueType>, "Build state-space model (CTMC or MA) for DFT", py::arg("dft"),
+          py::arg("symmetries") = storm::dft::storage::DftSymmetries(), py::arg("relevant_events") = storm::dft::utility::RelevantEvents(),
```
volkm, line 73: It is okay to add a default symmetry argument here.

Maybe just make it a separate change in the future such that there are no "hidden" changes which could be overlooked.

src/template_binding.h
volkm, line 1: Can you add some general documentation (maybe create `doc/developer.md`) on how to use the template bindings and maybe give a general example how to formulate such bindings? This would be very helpful as a guideline.

src/dft/analysis.cpp
```diff
@@ -51,8 +53,10 @@ void define_analysis(py::module& m) {
 }
 
 template<typename ValueType>
-void define_analysis_typed(py::module& m, std::string const& vt_suffix) {
-    py::classh<ExplicitDFTModelBuilder<ValueType>>(m, ("ExplicitDFTModelBuilder" + vt_suffix).c_str(), "Builder to generate explicit model from DFT")
+void define_analysis_typed(py::module& m) {
+    auto builder = stormpy::bindings::bindTemplateClass<ExplicitDFTModelBuilder<ValueType>, py::smart_holder>(
```
volkm: There is also a comment later on, but maybe add support for `classh` such that `smart_holder` does not need to be defined explicitly.

src/pars/model_instantiator.cpp
```diff
@@ -32,59 +34,40 @@ using MarkovAutomaton = storm::models::sparse::MarkovAutomaton<ValueType>;
...
-             "Instantiate model with given parameter values");
+template<storm::models::ModelType ModelKind, template<typename> class SparseModel, typename ValueType>
+void define_typed_instantiator(py::module& m) {
+    using ParametricModel = SparseModel<storm::RationalFunction>;
+    using InstantiatedModel = SparseModel<ValueType>;
+    using Instantiator = storm::utility::ModelInstantiator<ParametricModel, InstantiatedModel>;
+
+    auto implementation = stormpy::bindings::bindTemplateClass<Instantiator, py::smart_holder>(
+        m, "ModelInstantiator", stormpy::bindings::typeIndex<ModelKind, ValueType>(), "Instantiate a parametric model");
+    implementation.def(py::init<ParametricModel>(), "model"_a).def("instantiate", &Instantiator::instantiate, "Instantiate model with given parameter values");
```
volkm, line 45:
```suggestion
    implementation.def(py::init<ParametricModel>(), "model"_a)
        .def("instantiate", &Instantiator::instantiate, "Instantiate model with given parameter values");
```

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+    """
+
+    def deduction(family: TemplateClass, args: tuple[Any, ...], kwargs: Mapping[str, Any]) -> object:
+        if args:
+            instance = args[0]
+        elif keyword is not None and keyword in kwargs:
+            instance = kwargs[keyword]
+        else:
+            argument = f" {keyword!r}" if keyword is not None else " first positional"
+            raise TypeError(f"Cannot deduce template parameters without the{argument} argument")
```
volkm:
```suggestion
            raise TypeError(f"Cannot deduce template parameters without the {argument} argument")
```

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+    def canonical_name(self) -> str:
+        """Fully qualified public name of this template family."""
+        return self._canonical_name
+
+    @property
+    def metadata(self) -> TemplateMetadata:
+        """Return an immutable, current description for documentation tools."""
+        instantiations = tuple(
+            TemplateInstantiation(arguments, implementation, f"{implementation.__module__}.{implementation.__name__}")
+            for arguments, implementation in self._instantiations.items()
```
volkm, line 220:
```suggestion
            TemplateInstantiation(parameters, implementation, f"{implementation.__module__}.{implementation.__name__}")
            for parameters, implementation in self._instantiations.items()
```
To be consistent.

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+        :param implementation: Concrete Python class for those parameters.
+        :raises TypeError: If the parameter count is wrong or ``implementation``
+            is not a class.
+        :raises ValueError: If the parameter tuple is already registered.
+        """
+        key = self._normalize(parameters)
+        if not isinstance(implementation, type):
+            raise TypeError("A template implementation must be a class")
+        if key in self._instantiations:
+            raise ValueError(f"{self.__name__}{key!r} is already registered")
```
volkm, line 170:
```suggestion
            raise ValueError(f"{self.__name__} instantiation for {key!r} is already registered")
```

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def deduction(family: TemplateClass, args: tuple[Any, ...], kwargs: Mapping[str, Any]) -> object:
+        if args:
+            instance = args[0]
+        elif keyword is not None and keyword in kwargs:
+            instance = kwargs[keyword]
+        else:
+            argument = f" {keyword!r}" if keyword is not None else " first positional"
+            raise TypeError(f"Cannot deduce template parameters without the{argument} argument")
+        return (source or family).parameters_of(instance)
```
volkm, line 71: Maybe:
```suggestion
        return (source if source is not None else family).parameters_of(instance)
```

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+        self._deduction_guide = deduce
+        self._instantiations: dict[tuple[object, ...], type] = {}
+        for arguments, implementation in implementations.items():
+            self.register(arguments, implementation)
+
+    def _normalize(self, parameters: object) -> tuple[object, ...]:
+        """Convert subscription parameters to a validated tuple."""
+        key = parameters if isinstance(parameters, tuple) else (parameters,)
+        if len(key) != self._arity:
+            raise TypeError(f"{self.__name__} expects {self._arity} template " f"parameter{'s' if self._arity != 1 else ''}, got {len(key)}")
```
volkm, line 153:
```suggestion
            raise TypeError(f"{self.__name__} expects {self._arity} template parameter{'s' if self._arity != 1 else ''}, got {len(key)}")
```

lib/stormpy/_template.py
volkm, line 70:
```suggestion
            argument = f"{keyword!r}" if keyword is not None else " first positional"
            raise TypeError(f"Cannot deduce template parameters without the {argument} argument")
```

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+            instance = args[0]
+        elif keyword is not None and keyword in kwargs:
+            instance = kwargs[keyword]
+        else:
+            argument = f" {keyword!r}" if keyword is not None else " first positional"
+            raise TypeError(f"Cannot deduce template parameters without the{argument} argument")
+        return (source or family).parameters_of(instance)
+
+    deduction.__name__ = "deduce_from_first_argument"
+    deduction.__qualname__ = "deduce_from_first_argument"
```
volkm, line 74: Is it important here whether the first argument or keyword is used? And is the actual value relevant?

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+        self._parameters = parameter_metadata
+        self._arity = arity
+        self._deduction_guide = deduce
+        self._instantiations: dict[tuple[object, ...], type] = {}
+        for arguments, implementation in implementations.items():
+            self.register(arguments, implementation)
+
+    def _normalize(self, parameters: object) -> tuple[object, ...]:
+        """Convert subscription parameters to a validated tuple."""
+        key = parameters if isinstance(parameters, tuple) else (parameters,)
```
volkm, line 151: Claude mentions that, technically, using a tuple as value for a single parameter would behave the same as two parameters. But I would think this should not happen.

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+        self._instantiations[key] = implementation
+
+    def __getitem__(self, parameters: object) -> type:
+        """Return the concrete class registered for ``parameters``.
+
+        A single parameter may be written directly as ``family[T]``; multiple
+        parameters use normal subscription tuple syntax, ``family[T, U]``.
+
+        :raises TypeError: If the parameter count is wrong or no matching
+            specialization is registered.
```
volkm, line 180: Was it a deliberate choice not to raise a `KeyError` here?

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+        :raises TypeError: If the parameter count is wrong or ``implementation``
+            is not a class.
+        :raises ValueError: If the parameter tuple is already registered.
+        """
+        key = self._normalize(parameters)
+        if not isinstance(implementation, type):
+            raise TypeError("A template implementation must be a class")
+        if key in self._instantiations:
+            raise ValueError(f"{self.__name__}{key!r} is already registered")
+        self._instantiations[key] = implementation
```
volkm, line 171: Maybe add a check for inheritance by pair-wise comparison to already registered instantiations.
This avoids issues with deduction for inherited classes (see comment later on), but on the other-hand leads to quadratic runtime complexity when registering all instantiations.
Not sure whether this could be an issue, but for classes such as `SparseModel` and `Dtmc`, this could maybe happen.

```suggestion
        for other_key, other_implementation in self._instantiations.items():
            if issubclass(implementation, other_implementation) or issubclass(other_implementation, implementation):
                raise TypeError(f"{self.__name__} implementation {implementation!r} for {key!r} is related by inheritance to {other_implementation!r}, already registered for {other_key!r};")
        self._instantiations[key] = implementation
```

lib/stormpy/_template.py
```diff
@@ -0,0 +1,242 @@
...
+        parameters = self._deduction_guide(self, args, MappingProxyType(kwargs))
+        return self[parameters](*args, **kwargs)
+
+    def parameters_of(self, instance: object) -> tuple[object, ...]:
+        """Return the complete parameter tuple of a registered instance.
+
+        :raises TypeError: If the instance matches zero or multiple registered
+            specializations.
+        """
+        matches = [parameters for parameters, implementation in self._instantiations.items() if isinstance(instance, implementation)]
```
volkm, line 205: Claude mentions that `isinstance` also triggers for superclasses when using inheritance. So this could lead to issues when both the base class and the inherited class are instantiations. Maybe add the check in `register` proposed above.

lib/stormpy/dft/__init__.py
```diff
@@ -6,58 +6,71 @@
...
-    if isinstance(ft, DFT_double):
-        return _dft._analyze_dft_double(ft, properties, symred, allow_modularisation, relevant_events, allow_dc_for_relevant)
-    else:
-        assert isinstance(ft, DFT_ratfunc)
-        return _dft._analyze_dft_ratfunc(ft, properties, symred, allow_modularisation, relevant_events, allow_dc_for_relevant)
+DFT = TemplateClass(
+    "stormpy.dft.DFT",
+    _dft,
+    parameters=("ValueType",),
+    deduce=_deduce_from_first_argument(keyword="dft"),
```
volkm, line 18: The classes are defined as string, which is not easily type-checkable anymore. As far as I see, the keyword also has to match the name of the argument in the constructor. So this could lead to issues due to typos or changing argument names in the C++ side.
Is there any way to check the correspondence or do we just have to make sure to extensively test?

volkm (commented):
Overall, it looks really great. It is a very cool idea to integrate templating and will make bindings for templated classes so much easier and nicer.

I have a couple of comments, but mostly special cases.
I would suggest to add some guideline on how to properly write these templated bindings.

lib/stormpy/_template.py
volkm: We could also add a reverse mapping `self._parameters_by_type[implementation] = key` to the `register` function. Then this lookup could be performed via `self._parameters_by_type[type(instance)]` in O(1) instead O(n). (Depends a bit on the number of different implementations but for `Model`, it could make a difference.)

lib/stormpy/dft/simulator.py
```diff
@@ -104,11 +104,11 @@ def _update(self):
         for f in self._state.failable():
             if f.is_due_dependency():
                 self._is_failable_dependency = True
-                fail_dependency = f.as_dependency_double(self._dft)
+                fail_dependency = f.as_dependency(self._dft)
```
volkm, line 107: Note for myself: this will be costly if we need to deduce the underlying type for each simulation step. Better use direct access to the underlying method (double/ratfunc).
Is this possible?

examples/dfts/03-generic-types.py
linusheck: Names prefixed with an underscore will not import on `*`, which is what `dft` imports `_dft` with, so I think that this is really the correct path.

lib/stormpy/dft/simulator.py
linusheck: I'm not sure whether this is necessarily more costly, as this is just the native pybind11 type deduction. I will check this

examples/dfts/03-generic-types.py
volkm: Ah right, I did miss this part. Then please ignore my comment.

src/binding_type_index.h
linusheck: Indeed, thanks

src/template_binding.h
linusheck: Yes, before the caller had to use `smart_holder`, now we always use it :)

linusheck:
Build currently fails because of the `ExtendedNumber` mismatch. I won't fix this for now.

tests/dft/test_dft.py
linusheck: Done

lib/stormpy/_template.py
linusheck: As `[T]` is a type specialisation here, I think `TypeError` makes the most sense. But both are fine.

lib/stormpy/_template.py
linusheck: I like the reverse mapping. I think I'd just do this, and not use `isinstance` / `issubclass`. This solves the problem above too.

src/pars/model_instantiator.cpp
volkm: Okay, clang-format does not like it :)

lib/stormpy/_template.py
volkm: True, that should be good solution for both things.

lib/stormpy/_template.py
linusheck: I'm checking this with the map (mentioned below) now. Thus, I'm doing exact checks instead of using `isinstance`. That fixes the issue with inheritance.

lib/stormpy/dft/__init__.py
linusheck: `"stormpy.dft.DFT"` actually has to match the C++ name, or the statement will fail.

`ValueType` is just a Python-facing name, it doesn't have to match anything.

For `keyword=dft`, indeed, it is not checked whether this matches the `py::arg`, and I currently don't see how to do that, as pybind doesn't directly export this information. However, the keyword name `dft` is part of the public API so I don't see that getting renamed too often. Every existing caller of the API would also break. And I'd test the public API :)

volkm: Thanks for the clarification. So we should make sure to test all parts of the public-facing API and it should be fine.

lib/stormpy/dft/simulator.py
linusheck: Performance test:

```
import timeit

import stormpy.dft as d


for name, dft, direct_name in [
    (
        "double",
        d.load_dft_json_file("lib/stormpy/examples/files/dft/and.json"),
        "as_be_double",
    ),
    (
        "ratfunc",
        d.load_parametric_dft_galileo_file(
            "lib/stormpy/examples/files/dft/symmetry_param.dft"
        ),
        "as_be_ratfunc",
    ),
]:
    sim = d.DFTSimulator(
        dft,
        dft.state_generation_info(),
        d.RandomGenerator.create(42),
    )
    failable = next(iter(sim.get_state().failable()))
    generic = failable.as_be
    direct = getattr(failable, direct_name)
    for method_name, method in [("as_be", generic), (direct_name, direct)]:
        elapsed = timeit.timeit(lambda: method(dft), number=2_000_000)
        print(name, method_name, elapsed)
```

```
double as_be 0.6308111669932259
double as_be_double 0.5879930419905577
ratfunc as_be 1.7658001670060912
ratfunc as_be_ratfunc 0.9555176669964567
```

So it's slower especially on ratfunc, because that is the second registration. For this method, it's not an issue to just add back the specializations.

lib/stormpy/_template.py
linusheck: > Is it important here whether the first argument or keyword is used? 

No, it makes no difference. The value is irrelevant, only the type is relevant.

src/template_binding.h
linusheck: Yes, I'll do this as a part of a general notebook on how to write bindings. https://github.com/stormchecker/stormpy/issues/454

lib/stormpy/dft/simulator.py
volkm: Thanks for checking! Good to know that it indeed makes a difference, especially if the instantiation is not the first one.
I think we only have a few instance where such performance actually matters, so we can just provide the specialized functions there.

doc/source/doc/parametric_models.md
```diff
@@ -41,7 +41,7 @@ In order to obtain a standard DTMC, MDP or other Markov model, we need to instan
 ```{code-cell} python3
 import stormpy.pars
 
-instantiator = stormpy.pars.PDtmcInstantiator(model)
+instantiator = stormpy.pars.ModelInstantiator[stormpy.ModelType.DTMC, float](model)
```
sjunges, line 44: Very nitpicky. I like the generic way, but it is a bit annoying that we have to write stormpy.ModelType.DTMC rather than just DTMC or stormpy.DTMC.

examples/dfts/03-generic-types.py
```diff
@@ -0,0 +1,37 @@
+"""Demonstrate explicit and inferred stormpy template specializations."""
```
sjunges: Not sure that I would use examples for this... to some degree, we would want people to write the non-readable aliases. BUt I also dont know where else to put it. Maybe part of it is a test?

src/dft/dft.cpp
```diff
@@ -10,6 +10,7 @@
 #include <storm/adapters/RationalFunctionAdapter.h>
 #include <storm/settings/SettingsManager.h>
 
+#include "src/binding_type_index.h"
```
sjunges, line 13: should this be included in the helpers? We need it basically everywhere?

sjunges:
Great work, just a small number of nitpicks.

src/dft/dft.cpp
linusheck: ChatGPT complains that `helpers.h` is also used by pycarl, which doesn't need this.

examples/dfts/03-generic-types.py
linusheck: I agree, I moved a couple of lines from this to `test_dft`. I think this could be demonstrated in the notebook (https://github.com/stormchecker/stormpy/issues/454).

doc/source/doc/parametric_models.md
linusheck: This could be solved by aliasing these in stormpy:

```
DTMC = ModelType.DTMC
MDP = ModelType.MDP
POMDP = ModelType.POMDP
CTMC = ModelType.CTMC
MA = ModelType.MA
SMG = ModelType.SMG
```

Not sure if this is not more confusing? :D

linusheck:
I opened PRs for the different modules on my fork for now: https://github.com/linusheck/stormpy/pulls

Should we merge this first and then the modules? Or everything at once?

volkm:
Merging modules separately is the best way because it keeps the commits a bit smaller.

doc/source/doc/parametric_models.md
volkm: I would find aliasing more confusing, because for me `MA` would be the actual model and not the ModelType.

I would hope that this is only an issue on the developer side anyway. The user-facing API should deduce the types either automatically or we should provide some convenience functions.

PR #460, opened 2026-09-16
Move storage module to generic template bindings

linusheck:
Move storage to the new generic system (https://github.com/stormchecker/stormpy/pull/449).

All other modules depend on `storage`, so this should be merged first.


lib/stormpy/__init__.py
```diff
@@ -776,13 +776,13 @@ def eliminate_ECs(matrix, subsystem, possible_ecs, add_sink_row_states, add_self
...
         return _core._eliminate_end_components_RationalInterval(matrix, subsystem, possible_ecs, add_sink_row_states, add_self_loop_at_sink_states)
-    elif isinstance(matrix, storage.IntervalSparseMatrix):
+    elif isinstance(matrix, storage.SparseMatrix[stormpy.Interval]):
         return _core._eliminate_end_components_Interval(matrix, subsystem, possible_ecs, add_sink_row_states, add_self_loop_at_sink_states)
-    elif isinstance(matrix, storage.ExactSparseMatrix):
+    elif isinstance(matrix, storage.SparseMatrix[stormpy.Rational]):
         return _core._eliminate_end_components_Exact(matrix, subsystem, possible_ecs, add_sink_row_states, add_self_loop_at_sink_states)
-    elif isinstance(matrix, storage.SparseMatrix):
+    elif isinstance(matrix, storage.SparseMatrix[float]):
         return _core._eliminate_end_components_Double(matrix, subsystem, possible_ecs, add_sink_row_states, add_self_loop_at_sink_states)
```
volkm, line 786: This will be simplified after templating core as well, right?
Otherwise we could think about getting the valuetype out of the object such that we could maybe do a switch over the value type.

lib/stormpy/storage/__init__.py
```diff
@@ -2,6 +2,127 @@
...
+Add = TemplateClass(
+    "stormpy.storage.Add",
+    _storage,
+    parameters=(_TemplateParameter("DdType", kind="value"), "ValueType"),
+)
+AddIterator = TemplateClass(
+    "stormpy.storage.AddIterator",
+    _storage,
+    parameters=(_TemplateParameter("DdType", kind="value"), "ValueType"),
+)
```
volkm, line 48: Maybe we should group them by the underlying c++ file? Then it becomes easier to see where the classes come from and whether everything is covered.

lib/stormpy/storage/__init__.py
```diff
@@ -2,6 +2,127 @@
...
+    if args:
+        source = args[0]
+    else:
+        source = kwargs.get("components", kwargs.get("other_model"))
+        if source is None:
+            return (float,)
+    try:
+        return family.parameters_of(source)
+    except TypeError:
+        return SparseModelComponents.parameters_of(source)
```
volkm: This kind of function will be needed in slight variations a lot (see also `_deduce_from_model`). Can we create a general method `_deduce_from_object` which is similar to `deduce_from_first_argument`. But instead of directly returning the instance, we first apply the provided function `get_type` on the instance and then return the underlying type.

lib/stormpy/storage/__init__.py
```diff
@@ -2,6 +2,127 @@
...
+)
+
+
+def _deduce_sparse_model_components(family, args, kwargs):
+    if args:
+        return SparseMatrix.parameters_of(args[0])
+    transition_matrix = kwargs.get("transition_matrix")
+    if transition_matrix is not None:
+        return SparseMatrix.parameters_of(transition_matrix)
+    return (float,)
```
volkm: I think it is a nice idea to use float as a default. But in case `transition_matrix` can not be found (maybe due to a typo) it will silently fall back to float. Could this be risk?

lib/stormpy/storage/__init__.py
```diff
@@ -58,16 +179,7 @@ def get_maximal_end_components(model):
...
-        return stormpy.MaximalEndComponentDecomposition_ratfunc(model)
-    elif model.supports_uncertainty and model.is_exact:
-        return stormpy.MaximalEndComponentDecomposition_ratinterval(model)
-    elif model.is_exact:
-        return stormpy.MaximalEndComponentDecomposition_exact(model)
-    elif model.supports_uncertainty:
-        return stormpy.MaximalEndComponentDecomposition_interval(model)
-    else:
-        return stormpy.MaximalEndComponentDecomposition_double(model)
+    return MaximalEndComponentDecomposition(model)
```
volkm, line 178: So much nicer :-)

lib/stormpy/_template.py
```diff
@@ -51,6 +51,17 @@ class TemplateMetadata:
 DeductionGuide: TypeAlias = Callable[["TemplateClass", tuple[Any, ...], Mapping[str, Any]], object]
 
 
+def deduce_default(*parameters: object) -> DeductionGuide:
```
volkm, line 54: Also extend the documentation later on for this deduction.

tests/core/test_template.py
```diff
@@ -30,6 +30,14 @@ def test_deduction_selects_exact_registered_subclass():
     assert result.source is source
 
 
+def test_default_deduction_selects_configured_instantiation():
+    family = make_family(deduce=deduce_default("base"))
+
+    result = family()
+
+    assert type(result) is BaseImplementation
+
```
volkm, line 43: Can we add a test where the default does not exist?

src/storage/memorystructure.cpp
```diff
@@ -25,15 +27,16 @@ void define_memorystructure_untyped(py::module& m) {
...
     py::classh<storm::storage::SparseModelMemoryProductReverseData> memoryProductReverseData(m, "SparseModelMemoryProductReverseData");
-    define_memorystructure_product_each<double>(memoryStructure, memoryProductReverseData, "_double");
-    define_memorystructure_product_each<storm::RationalNumber>(memoryStructure, memoryProductReverseData, "_exact");
-    define_memorystructure_product_each<storm::RationalFunction>(memoryStructure, memoryProductReverseData, "_parametric");
+    define_memorystructure_product_each<double>(memoryStructure, memoryProductReverseData);
+    define_memorystructure_product_each<storm::RationalNumber>(memoryStructure, memoryProductReverseData);
+    define_memorystructure_product_each<storm::RationalFunction>(memoryStructure, memoryProductReverseData);
 }
 
 template<typename VT>
```
volkm:
```suggestion
template<typename ValueType>
```
Just for consistency

lib/stormpy/storage/__init__.py
```diff
@@ -264,36 +376,3 @@ def _get_integer_value(self, var):
...
-
-# Extend class MemoryStructure
-def product_model(self: MemoryStructure, model):
-    """
-    Compute the product of the memory structure with the model.
-    """
-    if model.supports_parameters:
-        return self._product_model_parametric(model)
-    elif model.supports_uncertainty:
-        raise NotImplementedError(f"product_model is not supported for interval models")
```
volkm, line 277: Might be related to the Copilot comment, but how is this now handled?

volkm (approved):
Great work! In general, it looks good to me.

lib/stormpy/__init__.py
linusheck: Yes, indeed, this will be simplified by templating core. I think getting the valuetype out of the object may be a good idea anyway, although the semantics of `isinstance` is slightly different than a `switch` I believe with inheritance (not that this would matter here)

lib/stormpy/storage/__init__.py
linusheck: Done :)

lib/stormpy/storage/__init__.py
linusheck: I agree we shouldn't have this default when the argument that is being deduced from is required anyway. In case of a typo, it will indeed deduce `float` but the construction will fail directly afterward because you passed an unsupported keyword.

lib/stormpy/_template.py
linusheck: Same for `deduce_from_object`

lib/stormpy/storage/__init__.py
linusheck: The C++ function `define_memorystructure_product_each` that binds `product_model` is simply not called for interval models.

tests/core/test_template.py
```diff
@@ -50,3 +65,60 @@ def test_cannot_register_implementation_for_multiple_parameters():
...
+def test_object_deduction_argument_selection():
+    guide = deduce_from_object(lambda obj: float if obj is None else obj.kind, keyword="model")
+    source = SimpleNamespace(kind=int)
+    alternate = SimpleNamespace(kind=str)
+
+    assert guide(None, (source,), {"model": alternate}) is int
+    assert guide(None, (), {"model": alternate}) is str
+    with pytest.raises(TypeError, match="missing argument.*model"):
+        guide(None, (), {})
+    assert guide(None, (), {"model": None}) is float
```
volkm:
```suggestion
    assert call_guide(guide, source, model=alternate) is int
    assert call_guide(guide, model=alternate) is str
    with pytest.raises(TypeError, match="missing argument.*model"):
        call_guide(guide)
    assert call_guide(guide, model=None) is float
```

tests/core/test_template.py
```diff
@@ -50,3 +65,60 @@ def test_cannot_register_implementation_for_multiple_parameters():
...
+
+
+def test_object_deduction_keyword_aliases():
+    guide = deduce_from_object(type, keyword=("components", "other_model"))
+
+    assert guide(None, (), {"other_model": 1}) is int
+    assert guide(None, (), {"components": 1.0, "other_model": 1}) is float
+    assert guide(None, (), {"components": None, "other_model": 1}) is type(None)
+    assert guide(None, (1,), {"components": 1.0}) is int
```
volkm:
```suggestion
    assert call_guide(guide, other_model=1) is int
    assert call_guide(guide, components=1.0, other_model=1) is float
    assert call_guide(guide, components=None, other_model=1) is type(None)
    assert call_guide(guide, 1, components=1.0) is int
```

tests/core/test_template.py
```diff
@@ -50,3 +65,60 @@ def test_cannot_register_implementation_for_multiple_parameters():
...
+
+
+def test_object_deduction_second_argument():
+    guide = deduce_from_object(type, keyword="model", position=1)
+
+    assert guide(None, (2, 1.0), {"model": 1}) is float
+    assert guide(None, (2,), {"model": 1.0}) is float
+    assert guide(None, (), {"model": 1.0}) is float
+    with pytest.raises(TypeError, match="missing argument.*position 1.*model"):
+        guide(None, (2,), {})
```
volkm:
```suggestion
    assert call_guide(guide, 2, 1.0, model=1) is float
    assert call_guide(guide, 2, model=1.0) is float
    assert call_guide(guide, model=1.0) is float
    with pytest.raises(TypeError, match="missing argument.*position 1.*model"):
        call_guide(guide, 2)
```

tests/core/test_template.py
```diff
@@ -50,3 +65,60 @@ def test_cannot_register_implementation_for_multiple_parameters():
...
+        family(soruce=SimpleNamespace(kind="base"))
+
+
+def test_object_deduction_explicit_default_only_applies_to_missing_argument():
+    guide = deduce_from_object(type, keyword="source", default=(float,))
+
+    assert guide(None, (), {}) == (float,)
+    assert guide(None, (), {"source": 1}) is int
+    assert guide(None, (None,), {}) is type(None)
+    assert guide(None, (), {"source": None}) is type(None)
```
volkm:
```suggestion
    assert call_guide(guide) == (float,)
    assert call_guide(guide, source=1) is int
    assert call_guide(guide, None) is type(None)
    assert call_guide(guide, source=None) is type(None)
```

tests/core/test_template.py
volkm, line 21:
```suggestion

def call_guide(guide, *args, **kwargs):
    return guide(None, args, kwargs)


```

volkm (commented):
It took me a while to properly understand the tests. I found using a helper function to mimic the actual call, made it easier for me.
All of the suggested changes should work as one large change (but I did not test it yet).

volkm:
Apart from maybe modifying the test, everything looks good to me. In particular the new `deduce_from_object` seems very helpful for all future bindings.

PR #461, opened 2026-09-16
added binding for discounted reward formula

TheGreatfpmK:
Not sure if this is how we want to handle the templating of getDiscountFactor method, but I didn't want to add the logic for templating since this is the only binding that needs in the formulae.cpp file


volkm:
Thanks for the bindings. The current form is okay, we are currently rewriting the templating anyway (based on https://github.com/stormchecker/stormpy/pull/449) and will touch this then.

Can you add a test as well to check that the bindings work as expected?

TheGreatfpmK:
Yeah I will add tests, I actually realized that I need the discountedTotalRewardFormula for me but I guess I can keep bindings for both

volkm (approved):
Other than the one Copilot comment, it looks good to me.

PR #463, opened 2026-09-18
Move core module to generic template bindings

linusheck:
Move core to the new generic system (https://github.com/stormchecker/stormpy/pull/449).


lib/stormpy/__init__.py
```diff
@@ -445,56 +497,23 @@ def check_model_sparse(model, property, only_initial_states=False, extract_sched
...
-            if hint:
-                task.set_hint(hint)
-            return _core._model_checking_sparse_engine(model, task, environment=environment)
+        if model.supports_parameters:
+            raise NotImplementedError("Model checking of partially observable models is not supported for parametric models.")
+        if model.supports_uncertainty:
+            raise NotImplementedError("Model checking of partially observable models is not supported for interval models.")
+    elif formula.is_multi_objective_formula and not model.supports_parameters:
+        return _core._multi_objective_model_checking(model, formula, environment=environment)
+    task = CheckTask[storage.parameters_of_model(model)](formula, only_initial_states)
```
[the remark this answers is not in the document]
linusheck: There should really be a test covering this :D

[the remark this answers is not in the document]
linusheck: As far as I understand, this isn't a regression compared to the older version, it just hits a different error now. I added an explicit error.

lib/stormpy/__init__.py
```diff
@@ -506,17 +527,10 @@ def check_model_dd(model, property, only_initial_states=False, environment=Envir
...
-    if model.supports_parameters:
-        task = _core.ParametricCheckTask(formula, only_initial_states)
-        return _core._parametric_model_checking_dd_engine(model, task, environment=environment)
-    else:
-        task = _core.CheckTask(formula, only_initial_states)
-        return _core._model_checking_dd_engine(model, task, environment=environment)
+    formula = property.raw_formula if isinstance(property, Property) else property
+    value_type = RationalFunction if model.supports_parameters else Rational if model.is_exact else float
+    task = CheckTask[value_type](formula, only_initial_states)
+    return _core._model_checking_dd_engine(model, task, environment=environment)
```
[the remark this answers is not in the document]
linusheck: Same here

lib/stormpy/__init__.py
```diff
@@ -663,12 +656,7 @@ def topological_sort(model, forward=True, initial=[]):
     :return: A topological sort of the states
     """
     matrix = model.transition_matrix if forward else model.backward_transition_matrix
-    if isinstance(model, storage.SparseModel[stormpy.RationalFunction]):
-        return storage._storage._topological_sort_rf(matrix, initial)
-    elif isinstance(model, storage.SparseModel[float]):
-        return storage._storage._topological_sort_double(matrix, initial)
-    else:
-        raise stormpy.exceptions.StormError("Unknown kind of model.")
+    return storage._storage._topological_sort(matrix, initial)
```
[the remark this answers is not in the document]
linusheck: Same type of issue

src/storage/dd.cpp
```diff
@@ -48,7 +48,7 @@ void define_dd_typed(py::module& m, py::classh<storm::dd::Dd<DdType>> const& dd)
 }
 
 void define_dd_nt(py::module& m) {
-    py::native_enum<storm::dd::DdType>(m, "DdType", "enum.Enum").value("Sylvan", storm::dd::DdType::Sylvan).value("CUDD", storm::dd::DdType::CUDD).finalize();
```
volkm, line 51: Given that `storm::dd::DdType` is (currently) defined in `storm/storage/dd/DdType.h` I would do it the other way: keep the definition here and import it in core.

src/storage/matrix.cpp
```diff
@@ -22,11 +22,11 @@ using row_index = unsigned int;
...
-        "_topological_sort_double",
+        "_topological_sort",
         [](SparseMatrix<double>& matrix, std::vector<uint64_t> initial) { return storm::utility::graph::getTopologicalSort(matrix, initial); }, "matrix"_a,
         "initial"_a, "get topological sort w.r.t. a transition matrix");
     m.def(
-        "_topological_sort_rf",
+        "_topological_sort",
         [](SparseMatrix<storm::RationalFunction>& matrix, std::vector<uint64_t> initial) { return storm::utility::graph::getTopologicalSort(matrix, initial); },
         "matrix"_a, "initial"_a, "get topological sort w.r.t. a transition matrix");
 }
```
volkm: I think this can be part of a templated definition in `define_sparse_matrix()`

src/core/core.cpp
```diff
@@ -174,7 +171,8 @@ void define_build_sparse_model_defs(py::module& m) {
         m.def("make_sparse_model_builder_parametric", &storm::api::makeExplicitModelBuilder<storm::RationalFunction>, "Construct a builder instance",
               py::arg("model_description"), py::arg("options"), py::arg("action_mask") = nullptr,
               py::arg("exploration_options") = typename storm::builder::ExplicitModelBuilder<ValueType>::Options());
-        py::classh<storm::builder::ExplicitModelBuilder<storm::RationalFunction>>(m, "ExplicitParametricModelBuilder", "Model builder for sparse models")
+        stormpy::bindings::bindTemplateClass<storm::builder::ExplicitModelBuilder<storm::RationalFunction>>(
+            m, "ExplicitModelBuilder", stormpy::bindings::typeIndex<storm::RationalFunction>(), "Model builder for sparse models")
             .def("build", &storm::builder::ExplicitModelBuilder<storm::RationalFunction>::build, "Build the model", py::call_guard<py::gil_scoped_release>())
             .def("export_lookup", &storm::builder::ExplicitModelBuilder<storm::RationalFunction>::exportExplicitStateLookup, "Export a lookup model");
     } else if constexpr (std::is_same_v<ValueType, storm::RationalNumber>) {
```
volkm: I think some of these function could be merged as well. I am not sure whether we need the branching on the ValueType anymore.

src/core/core.cpp
```diff
@@ -270,26 +270,7 @@ void exportDRN(std::shared_ptr<storm::models::sparse::Model<ValueType>> model, s
...
-        prefix = "";
-        suffix = "_interval";
-    } else if constexpr (std::is_same_v<ValueType, storm::RationalInterval>) {
-        prefix = "_exact";
-        suffix = "_interval";
-    }
-
-    m.def(("_export" + prefix + "_to_drn" + suffix).c_str(), &exportDRN<ValueType>,
+    m.def("_export_to_drn", &exportDRN<ValueType>,
           ("Export " + (std::is_same_v<ValueType, storm::RationalFunction> ? std::string("parametric ") : std::string()) + "model in DRN format").c_str(),
```
volkm:
```suggestion
          "Export model in DRN format",
```

src/core/simulator.cpp
```diff
@@ -4,13 +4,15 @@
...
+
 template<typename ValueType>
 using PLSim = storm::simulator::DiscreteTimePrismProgramSimulator<ValueType>;
 
 template<typename ValueType>
-void define_sparse_model_simulator(py::module& m, std::string const& vtSuffix) {
-    py::classh<storm::simulator::DiscreteTimeSparseModelSimulator<ValueType>> dtsmsd(m, ("_DiscreteTimeSparseModelSimulator" + vtSuffix).c_str(),
-                                                                                     "Simulator for sparse discrete-time models in memory (for ValueType)");
+void define_sparse_model_simulator(py::module& m) {
+    auto dtsmsd = stormpy::bindings::bindTemplateClass<storm::simulator::DiscreteTimeSparseModelSimulator<ValueType>>(
```
volkm, line 14: I have no idea what the "d" in "dtsmsd" stands for but it was already there before :)

src/core/transformation.cpp
volkm: Was not your scope, but can you move these instantiations to the right place? I think the definitions in `define_transformation_mdef` should just be part of `define_transformation_typed`.

src/core/modelchecking.cpp
volkm: Again, not your scpe, but move the definitions to something like `define_modelchecking_typed`?

src/core/modelchecking.cpp
```diff
@@ -156,59 +156,35 @@ void define_check_task(py::module& m, std::string const& name) {
...
+    if constexpr (std::is_same_v<ValueType, double> || std::is_same_v<ValueType, storm::RationalNumber>) {
+        m.def("_compute_expected_number_of_visits", &getExpectedNumberOfVisits<ValueType>, py::arg("env"), py::arg("model"));
+        m.def("_compute_steady_state_distribution", &getSteadyStateDistribution<ValueType>, py::arg("env"), py::arg("model"));
+        m.def("_model_checking_fully_observable", &modelCheckingFullyObservableSparseEngine<ValueType>, py::arg("model"), py::arg("task"),
               py::arg("environment"));
-        m.def("_exact_model_checking_sparse_engine", &modelCheckingSparseEngine<storm::RationalNumber>, "Perform model checking using the sparse engine",
+        m.def("_multi_objective_model_checking", &multiObjectiveModelChecking<ValueType>, "Run multi-objective model checking", py::arg("model"),
+              py::arg("formula"), py::arg("environment"));
+    }
+    if constexpr (std::is_same_v<ValueType, double> || std::is_same_v<ValueType, storm::RationalFunction>) {
```
volkm: Are these functions not defined for RationalNumber? I see that the current versions had no bindings, but I think they should be supported.

lib/stormpy/__init__.py
```diff
@@ -7,6 +7,64 @@
...
+    TemplateParameter as _TemplateParameter,
+    deduce_default as _deduce_default,
+    deduce_from_object as _deduce_from_object,
+)
+
+# src/core/core.cpp
+ExplicitModelBuilderOptions = _TemplateClass("stormpy.ExplicitModelBuilderOptions", _core, parameters=("ValueType",), deduce=_deduce_default(float))
+ExplicitModelBuilder = _TemplateClass("stormpy.ExplicitModelBuilder", _core, parameters=("ValueType",))
+ActionMask = _TemplateClass("stormpy.ActionMask", _core, parameters=("ValueType",))
+StateValuationFunctionActionMask = _TemplateClass("stormpy.StateValuationFunctionActionMask", _core, parameters=("ValueType",), deduce=_deduce_default(float))
```
volkm, line 21: Is there a rule when to use `_deduce_default` and when not?

lib/stormpy/__init__.py
```diff
@@ -506,17 +527,12 @@ def check_model_dd(model, property, only_initial_states=False, environment=Envir
...
-    if model.supports_parameters:
-        task = _core.ParametricCheckTask(formula, only_initial_states)
-        return _core._parametric_model_checking_dd_engine(model, task, environment=environment)
-    else:
-        task = _core.CheckTask(formula, only_initial_states)
-        return _core._model_checking_dd_engine(model, task, environment=environment)
+    if model.is_exact and not model.supports_parameters:
+        raise NotImplementedError("Model checking of exact symbolic models is not supported by the dd engine.")
+    formula = property.raw_formula if isinstance(property, Property) else property
+    value_type = RationalFunction if model.supports_parameters else Rational if model.is_exact else float
```
volkm:
```suggestion
    value_type = RationalFunction if model.supports_parameters else (Rational if model.is_exact else float)
```
Just to make the order clear

lib/stormpy/__init__.py
```diff
@@ -506,17 +527,12 @@ def check_model_dd(model, property, only_initial_states=False, environment=Envir
...
-        formula = property
-
-    if model.supports_parameters:
-        task = _core.ParametricCheckTask(formula, only_initial_states)
-        return _core._parametric_model_checking_dd_engine(model, task, environment=environment)
-    else:
-        task = _core.CheckTask(formula, only_initial_states)
-        return _core._model_checking_dd_engine(model, task, environment=environment)
+    if model.is_exact and not model.supports_parameters:
+        raise NotImplementedError("Model checking of exact symbolic models is not supported by the dd engine.")
```
volkm: It should be supported, but is missing the bindings. See comments in `core/modelchecking.cpp`.

lib/stormpy/__init__.py
```diff
@@ -528,17 +544,12 @@ def check_model_hybrid(model, property, only_initial_states=False, environment=E
...
-    if model.supports_parameters:
-        task = _core.ParametricCheckTask(formula, only_initial_states)
-        return _core._parametric_model_checking_hybrid_engine(model, task, environment=environment)
-    else:
-        task = _core.CheckTask(formula, only_initial_states)
-        return _core._model_checking_hybrid_engine(model, task, environment=environment)
+    if model.is_exact and not model.supports_parameters:
+        raise NotImplementedError("Model checking of exact symbolic models is not supported by the hybrid engine.")
+    formula = property.raw_formula if isinstance(property, Property) else property
+    value_type = RationalFunction if model.supports_parameters else Rational if model.is_exact else float
```
volkm:
```suggestion
    value_type = RationalFunction if model.supports_parameters else (Rational if model.is_exact else float)
```

lib/stormpy/__init__.py
```diff
@@ -528,17 +544,12 @@ def check_model_hybrid(model, property, only_initial_states=False, environment=E
...
-        formula = property
-
-    if model.supports_parameters:
-        task = _core.ParametricCheckTask(formula, only_initial_states)
-        return _core._parametric_model_checking_hybrid_engine(model, task, environment=environment)
-    else:
-        task = _core.CheckTask(formula, only_initial_states)
-        return _core._model_checking_hybrid_engine(model, task, environment=environment)
+    if model.is_exact and not model.supports_parameters:
+        raise NotImplementedError("Model checking of exact symbolic models is not supported by the hybrid engine.")
```
volkm:
```suggestion
```
I think this could be supported.

lib/stormpy/__init__.py
```diff
@@ -662,13 +655,14 @@ def topological_sort(model, forward=True, initial=[]):
     :param initial: a list of states
     :return: A topological sort of the states
     """
+    if not model.is_sparse_model:
+        raise NotImplementedError("Topological sorting is only supported for sparse models.")
+    if model.supports_uncertainty:
+        raise NotImplementedError("Topological sorting of interval models is not supported.")
+    if model.is_exact and not model.supports_parameters:
+        raise NotImplementedError("Topological sorting of exact models is not supported.")
```
volkm: Maybe is supported?

tests/core/test_core_templates.py
```diff
@@ -0,0 +1,64 @@
...
+    program = stormpy.parse_prism_program(get_example_path("dtmc", "die.pm"))
+    model = stormpy.build_symbolic_model(program)
+    result = stormpy.create_filter_initial_states_symbolic(model)
+
+    assert stormpy.DdType is stormpy.storage.DdType
+    assert type(result) is stormpy.SymbolicQualitativeCheckResult[stormpy.DdType.Sylvan]
+    assert stormpy.Bdd.parameters_of(result.get_truth_values()) == (stormpy.DdType.Sylvan,)
+
+
+@pytest.mark.parametrize("exact", [False, True])
```
volkm: Parametrize with the value type? Then it is more flexible.

tests/core/test_core_templates.py
volkm, line 1: All these functions test functionality of different files in `src/core`. Can we move them to corresponding tests in `tests`, for example in `core`, in `logic` or in `storage`?

tests/core/test_core_templates.py
```diff
@@ -0,0 +1,64 @@
+import pytest
+
+import stormpy
+from helpers.helper import get_example_path
+
+
+@pytest.mark.parametrize("value_type", [float, stormpy.Rational, stormpy.RationalFunction, stormpy.Interval, stormpy.RationalInterval])
```
volkm: This is a good idea and we should probably do it for more tests in the future.

tests/core/test_core_templates.py
```diff
@@ -0,0 +1,64 @@
...
+    program = stormpy.parse_prism_program(get_example_path("dtmc", "die.pm"))
+    if value_type in (stormpy.Rational, stormpy.RationalInterval):
+        model = stormpy.build_sparse_exact_model(program)
+    elif value_type is stormpy.RationalFunction:
+        model = stormpy.build_parametric_model(program)
+    else:
+        model = stormpy.build_model(program)
+    if value_type in (stormpy.Interval, stormpy.RationalInterval):
+        uncertainty = 0.01 if value_type is stormpy.Interval else stormpy.Rational("1/100")
+        model = stormpy.AddUncertainty(model).transform(uncertainty)
```
volkm: This seems a good helper function in general. Maybe add a function `build_sparse_model` which given a path to a prism program and a valuetype, builds the right model?

volkm (commented):
In general, it looks great. Thanks for the effort.
I have some comments, for example I think we could also add bindings for exact symbolic models.

src/storage/matrix.cpp
linusheck: Yes, that simplifies it a lot :)

src/core/core.cpp
linusheck: it seems that `build_sparse_model_from_explicit` really only works for `double`, but everything else can indeed be generic

lib/stormpy/__init__.py
linusheck: yes :)

lib/stormpy/__init__.py
linusheck: yes :)

lib/stormpy/__init__.py
linusheck: yes :)

src/core/modelchecking.cpp
linusheck: yes :)

lib/stormpy/__init__.py
linusheck: I think that we should come up with one.

Maybe if the following apply:

- there is a constructor that is expected to be used by users and does not have any input to infer the type from,
- there is no reason to believe why it would be dangerous or confusing to infer the type

